PR #9, opened 2022-12-24
WIP: Add nix based packaging

DrRuhe:
Hello,

This PR mainly adds files required for easily installing samply under [Nix](https://nixos.org/).
The main two files needed for nix packaging to work would be the `flake.nix` and the `flake.lock`. A `Cargo.lock` is also required, but since samply is an application and not a library I would assume this is ok. 

I mainly want to know if there would be interest in merging something like this, at which point I would do some more cleanup.

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,1 @@
+use_flake
--- a/.github/workflows/nix-packaging.yml
+++ b/.github/workflows/nix-packaging.yml
@@ -1,0 +1,33 @@
+# This is a basic workflow to help you get started with Actions
+
+name: "Nix Packaging"
+
+# Controls when the workflow will run
+on:
+  # Triggers the workflow on push or pull request events but only for the "main" branch
+  push:
+    branches: [ "main" ]
+  pull_request:
+    branches: [ "main" ]
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+jobs:
+  build:
+    name: "Build Samply using nix flake"
+    runs-on: ubuntu-22.04
+    timeout-minutes: 30
+    steps:
+      - uses: actions/checkout@v3
+      - uses: cachix/install-nix-action@v17
+        with:
+          nix_path: nixpkgs=channel:nixos-22.05
+      # - uses: cachix/cachix-action@v10
+      #   with:
+      #     name: samply
+      #     authToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
+
+      - name: Build samply with nix
+        run: nix build -L .#
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,9 @@
 target
 profile.json
 .DS_Store
+.direnv/
+
+result-bin
 
 # These libxul blobs are too big to put them in the repo.
 # The benchmarks example downloads them from the Mozilla symbol server on demand.
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,164 @@
+{
+  "nodes": {
+    "advisory-db": {
+      "flake": false,
+      "locked": {
+        "lastModified": 1674318054,
+        "narHash": "sha256-6ohOG3PwPhxNaAXooe+kZKkBUCcGfxylWlTcdAI12Jg=",
+        "owner": "rustsec",
+        "repo": "advisory-db",
+        "rev": "bed742d611e93f1b66fb8accf224857b26e02ebf",
+        "type": "github"
+      },
+      "original": {
+        "owner": "rustsec",
+        "repo": "advisory-db",
+        "type": "github"
+      }
+    },
+    "crane": {
+      "inputs": {
+        "flake-compat": "flake-compat",
+        "flake-utils": "flake-utils",
+        "nixpkgs": [
+          "nixpkgs"
+        ],
+        "rust-overlay": "rust-overlay"
+      },
+      "locked": {
+        "lastModified": 1674348649,
+        "narHash": "sha256-hBRlaUlsrmW1wAPevwQnkrT0XiLrmlAHWabWYmLeQlQ=",
+        "owner": "ipetkov",
+        "repo": "crane",
+        "rev": "ccea7b33178daf6010aae3ea2b3fb5b0241b9146",
+        "type": "github"
+      },
+      "original": {
+        "owner": "ipetkov",
+        "repo": "crane",
+        "type": "github"
+      }
+    },
+    "flake-compat": {
+      "flake": false,
+      "locked": {
+        "lastModified": 1668681692,
+        "narHash": "sha256-Ht91NGdewz8IQLtWZ9LCeNXMSXHUss+9COoqu6JLmXU=",
+        "owner": "edolstra",
+        "repo": "flake-compat",
+        "rev": "009399224d5e398d03b22badca40a37ac85412a1",
+        "type": "github"
+      },
+      "original": {
+        "owner": "edolstra",
+        "repo": "flake-compat",
+        "type": "github"
+      }
+    },
+    "flake-utils": {
+      "locked": {
+        "lastModified": 1667395993,
+        "narHash": "sha256-nuEHfE/LcWyuSWnS8t12N1wc105Qtau+/OdUAjtQ0rA=",
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "rev": "5aed5285a952e0b949eb3ba02c12fa4fcfef535f",
+        "type": "github"
+      },
+      "original": {
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "type": "github"
+      }
+    },
+    "flake-utils_2": {
+      "locked": {
+        "lastModified": 1667395993,
+        "narHash": "sha256-nuEHfE/LcWyuSWnS8t12N1wc105Qtau+/OdUAjtQ0rA=",
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "rev": "5aed5285a952e0b949eb3ba02c12fa4fcfef535f",
+        "type": "github"
+      },
+      "original": {
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "type": "github"
+      }
+    },
+    "nixpkgs": {
+      "locked": {
+        "lastModified": 1674236650,
+        "narHash": "sha256-B4GKL1YdJnII6DQNNJ4wDW1ySJVx2suB1h/v4Ql8J0Q=",
+        "owner": "NixOS",
+        "repo": "nixpkgs",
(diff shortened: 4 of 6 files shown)


mstange:
Hi, I know almost nothing about Nix but I'm happy to accept changes that make it easier for people to install samply in their various configurations, as long as those changes don't come with an undue maintenance cost or impose certain limitations. For example, I want to retain the ability to require up-to-date Rust versions.

I looked at the patch and noticed that it introduces a cargo workspace. Does this simplify things?

Also, at the moment, if I make local changes to, say, samply-symbols, those are immediately picked up by local builds of the dependent crates in this repo (e.g. wholesym), even if I don't publish a new version of samply-symbols. Would such local changes cause the checksums in Cargo.lock to change?

Oh, I just remembered a potential pain point with a checked-in Cargo.toml: If I use `cargo release` to publish a new version of one of the crates, there's a "clean git repo status" check at some point during the publishing workflow; if I have VS Code with rust-analyzer open at the same time, the updating of the version in Cargo.toml (which happens as part of the publish) causes rust-analyzer to generate a new Cargo.lock immediately in the middle of publishing, and then the "clean repo" check fails. I avoided this problem by putting Cargo.lock on the gitignore. But maybe there is a better workaround for this.

DrRuhe:
>  I know almost nothing about Nix

No worries! In short, it is a declarative build system that can produce deterministic results. For people not using nix, the only noticeable difference is the existence of the `flake.nix/.lock` files. The other changes are just there to make packaging easier/possible. 

>  don't come with an undue maintenance cost or impose certain limitations

This packaging solution does not impose restrictions on the project itself. The only maintenance needed would be to update system dependencies when these change so that these are present in the build environment (this should happen very rarely, if at all. Currently this is just [openssl](https://github.com/DrRuhe/samply/blob/main/flake.nix#L55-L59)). 

> I looked at the patch and noticed that it introduces a cargo workspace. Does this simplify things?

Yes, since it allows `cargo build` to run in the root directory. Since the workspace makes heavy use of local dependencies, it would be difficult to include those folders in the build environment. It would make the packaging a whole lot more complicated. 

> Would such local changes cause the checksums in Cargo.lock to change?

Interesting Question! As it turns out, the cargo.lock does not include checksums (or the source registry) for local dependencies (here is the entry for [samply-symbols](https://github.com/DrRuhe/samply/blob/main/Cargo.lock#L1821-L1851), and an [entry of a dependency](https://github.com/DrRuhe/samply/blob/main/Cargo.lock#L2229-L2245)). So the Cargo.lock will stay the same regardless what local dependency changes happen.

> a potential pain point with a checked-in Cargo.toml

I assume you mean the `Cargo.lock`?
I have never released via `cargo release` myself, so I am unaware of the process details. So If I understand correctly:

- cargo release updates the version number
- rust-analyzer updates Cargo.lock to reflect the changes
- git repo is no longer clean -> check failed

Is it possible to do a "Release commit" which increments the version number and updates the lockfile and then use `cargo release`? In any case, I think there should be a way to circumvent this.

mstange:
> The only maintenance needed would be to update system dependencies when these change so that these are present in the build environment (this should happen very rarely, if at all. Currently this is just [openssl](https://github.com/DrRuhe/samply/blob/main/flake.nix#L55-L59)).

That sounds great. In fact I'd like samply to have zero system dependencies, and I thought I had eliminated the openssl dependency with https://github.com/mstange/symsrv/commit/527681170d4290ed6260112e8a65ef2bf400e3c6 - do you know what's dragging in the openssl dependency now? `cargo tree` under the `samply` directory doesn't seem to show it for me on macOS or Linux.

Oh, I think I know what it is: The `benchmarks` crate still drags in an openssl dependency, and it's now part of the workspace.

Of course we can change how `benchmarks` depends on `reqwest` to avoid this. But we may hit similar issues in the future, if I add further utilities that are "external" to the main samply crate, if those utilities require extra dependencies. It seems wasteful to have those utilities bloat the nix build, because they're not needed for the main binary.

> > I looked at the patch and noticed that it introduces a cargo workspace. Does this simplify things?
> 
> Yes, since it allows `cargo build` to run in the root directory. Since the workspace makes heavy use of local dependencies, it would be difficult to include those folders in the build environment. It would make the packaging a whole lot more complicated.

I see. However, the local dependencies are only to simplify development. For released versions, everything is on crates.io. For example, `cargo install samply` pulls the `samply-symbols` dependency from crates.io, not from the local directory. This is a requirement from crates.io: You can only publish a package to crates.io if all dependencies are also published on crates.io. So the "dependent crates" in this repo are no different from any other dependencies.

So, if there were a way to run `cargo build` in a "pull all dependencies from crates.io" mode, then I think we could add the nix script only to the `samply` crate and ignore the other directories. Unfortunately, I don't think there is such a mode. `cargo publish` pulls this off by creating a new, "normalized" `Cargo.toml` file which has all path dependency information stripped. You can download https://crates.io/api/v1/crates/samply/0.10.1/download to see an example of such a "normalized" file; give the downloaded file a `.tar.gz` extension and you'll see both a `Cargo.toml` and a `Cargo.toml.orig` file in the uncompressed directory.

Hmm. It looks like avoiding the workspace is a bit more trouble than I expected. I guess we could have the nix build create a "normalized" Cargo.toml and then pass that to `cargo build --manifest`, but at that point the workspace is probably the lesser evil.

> > Would such local changes cause the checksums in Cargo.lock to change?
> 
> Interesting Question! As it turns out, the cargo.lock does not include checksums (or the source registry) for local dependencies (here is the entry for [samply-symbols](https://github.com/DrRuhe/samply/blob/main/Cargo.lock#L1821-L1851), and an [entry of a dependency](https://github.com/DrRuhe/samply/blob/main/Cargo.lock#L2229-L2245)). So the Cargo.lock will stay the same regardless what local dependency changes happen.

Ok, good. Thanks for checking!

> > a potential pain point with a checked-in Cargo.toml
> 
> I assume you mean the `Cargo.lock`?

Er, yes.

> I have never released via `cargo release` myself, so I am unaware of the process details. So If I understand correctly:
> 
>     * cargo release updates the version number
> 
>     * rust-analyzer updates Cargo.lock to reflect the changes
> 
>     * git repo is no longer clean -> check failed

That's right.

> Is it possible to do a "Release commit" which increments the version number and updates the lockfile and then use `cargo release`? In any case, I think there should be a way to circumvent this.

I'm not sure, but I agree, there should be a way to get around this. Anyway, I'm fine with committing `Cargo.lock` for the main samply crate; doing so is best practice for a binary crate anyway. And if we do end up with a workspace, then I am also fine with committing the `Cargo.lock` of the workspace.

---

Two more things that came to mind:

 1. At some point in the future I'd like to add another binary target, `cargo-samply`, for the `samply` crate, so that two binaries get installed by `cargo install samply`. This is similar to how `cargo install flamegraph` installs the two binaries `flamegraph` and `cargo-flamegraph`. I'm expecting that this will require changes to the nix build script.
 2. We'll need a test that's run by github CI to verify that the nix build works. Otherwise I'm sure I'll break it eventually, by accident.

DrRuhe:
>  The benchmarks crate still drags in an openssl dependency

Yup, `benchmarks` it is.
```
benchmarks v0.1.0 (./tools/benchmarks)
├── reqwest v0.11.13
│   ├── hyper-tls v0.5.0
│   │   ├── native-tls v0.2.11
│   │   │   ├── openssl v0.10.45
│   │   │   ├── openssl-probe v0.1.5
│   │   │   └── openssl-sys v0.9.80 (*)
```
I think the culprit might be the `craneLib.buildDepsOnly` that AFAIK builds all the external dependencies that the workspace needs (thereby including benchmarks related things).
I don't know if this can be customized, [but I've asked](https://github.com/ipetkov/crane/discussions/204).

However I don't think that Its a big deal, for the following reasons:
- nix would manage the openssl installation for you, so "it just works"
- openssl is everywhere, so little to no extra disk space is required
- considerably more effort would have to go customizing the nix build to these demands (which would probably add undue maintenance cost)

> For example, cargo install samply pulls the samply-symbols dependency from crates.io, not from the local directory.

In that sense nix is different from running `cargo install samply`. You can think of it more like:
```
git clone <samply>
cargo install . --bin samply #note that I dont know if this would run but it conveys the point
```

It is not pullling samply from crates.io but rather building from source. Similar to above, changing it to behave like `cargo install samply` would require considerable effort and (definitely) introduce undue maintenance cost. 

Side-note:
Since nix is declarative and values reproducability, downloading things from the internet requires predetermined checksums. Since `cargo install samply` gets the currently latest version from `crates.io` recreating this behavior in nix would require at least one checksum to be maintained (manually). 

> At some point in the future I'd like to add another binary target, cargo-samply, for the samply crate

I think the current setup would just continue to work, but just install `samply`. In general, adding more binaries to be installed would require (minor) changes to the `flake.nix`.

> We'll need a test that's run by github CI to verify that the nix build works

About that. The current packaaging solution is based on [a crane template](https://github.com/ipetkov/crane/). Currently included with that are the checks:

- cargo-fmt: checks formatting
- cargo-doc: checks that the docs build (and I believe runs doctests, not sure)
- cargo-clippy: runs cargo clippy
- cargo-audit: checking dependencies for potential vulnerabilities.
- cargo-nextest: running the tests with nextest
- on x86_linux: cargo-coverage: code coverage measurements using cargo tarpaulin

These can all be run using `nix flake check -L .#` (the resulting logs have potential for improvement). But since there already is CI, we could just disable/remove (some of) them. Adding this to CI would also require some effort to add caching, otherwise it would build every dependecy every time, quickly eating up  valuable CI time. I have yet to try and use CI caching for nix with GitHub.

So, I guess the open question is which checks to keep and should CI be set up with caching?

mstange:
I've gone ahead and done the conversion to a workspace, and checked in the Cargo.lock. I also cleaned up some dependencies of the tools to make cargo-audit pass, and added cargo-clippy and cargo-fmt to the tests run in CI.

I don't have a strong opinion on the list of checks to run for the nix build, or whether to use caching. The project isn't all that big yet, so I think we should be fine as far as CI time is concerned.

DrRuhe:
Nice. I have disabled the checks in the flake in 994d4128f123ee6526efbf215e8b42d02685e12a to reduce the required CI time. I have also added a Nix Packaging action that should "just work". 

If it turns out it requires too much time to run the nix-based packaging in CI I would suggest caching with [Cachix](https://nix.dev/tutorials/continuous-integration-github-actions). Their free tier should be plenty and it integrates nicely with Github Actions.

mstange:
@DrRuhe What would be the next step here? Is this usable as-is, or should it be included in the nixpkgs registry?

DrRuhe:
Its usable already. Although upstreaming the packaging to nixpkgs would be the the ideal solution.

However I have yet to contribute to nixpkgs and I don't know how much work is required there. In particular, I don't think the current packaging approach can be upstreamed, so more work would be required. 

When I have more time available again, I might take a closer look.

mstange:
Ah ok, thanks

PR #10, opened 2022-12-31
Decompress section data prior to use

bobrik:
This doesn't matter for `.eh_frame`, but it does matter to dwarf data, which is often compressed. Accessing uncompressed data will result in a silent failure in `framehop`:

* https://github.com/mstange/framehop/blob/v0.7.2/src/unwinder.rs#L575


mstange:
Thanks! Yeah, the silent discarding of the error in framehop is not great.

PR #18, opened 2023-01-28
Add support for attaching profiler to a PID (Linux only). 

ishitatsuyuki:
This adds support for attaching to a PID by `samply record -p 12345`.

I only implemented (and feature flagged) this for Linux, since the attaching story seems more complicated on the macOS side.


mstange (approved):
Excellent, thanks! I'll merge this and then fix the macOS build. I should probably set up some macOS CI.

PR #21, opened 2023-02-02
AVMA calculation fixes and PE correlation hack for Wine on Linux.

ishitatsuyuki:
This PR contains a few fixes and tweaks to base AVMA calculation, and adds a PE mapping correlation hack to allow PE binaries to be traced back to their path which is otherwise not straightforward to determine from the memory mapping table.

I did a few basic tests and it seems nothing is breaking for the Linux side of stuff; I've also been running this with my PoC branch with SEH unwinding (https://github.com/ishitatsuyuki/samply/tree/poc, https://github.com/ishitatsuyuki/framehop/tree/pe-poc) which also works. The upstreaming of SEH unwinding will take a bit more though, since there are several spots to clean-up and some design decisions to discuss in detail.

See commit messages for detailed descriptions.


samply/src/linux_shared/mod.rs
```diff
@@ -1203,9 +1203,14 @@ where
...
-            Some(first_segment) => first_segment.address(),
-            None => 0,
+        let is_pe = file.format() == object::BinaryFormat::Pe;
+        let base_svma = if is_pe {
+            file.relative_address_base()
+        } else {
+            match file.segments().next() {
+                Some(first_segment) => first_segment.address(),
+                None => 0,
+            }
```
mstange, line 1321: It might be better to just call into `samply_symbols::relative_address_base` here. That way we'll know that whatever we do here is consistent with what symbolication does.

samply/src/linux_shared/mod.rs
```diff
@@ -399,7 +417,54 @@ where
         }
     }
 
+    /// This is a terrible hack to get binary correlation working with apps on Wine.
+    ///
+    /// Unlike ELF, PE has the notion of "file alignment" that is different from page alignment.
+    /// Hence, even if the virtual address is page aligned, its on-disk offset may not be. This
+    /// leads to obvious trouble with using mmap, since mmap requires the file offset to be page
+    /// aligned. Wine's workaround is straightforward: for misaligned sections, Wine will simply
+    /// copy the image from disk instead of mmaping them.
```
mstange, line 426: Oh, but it also mmaps them first, otherwise this code wouldn't be called, right? I think this would be good to call out here. Reading the comment I was thinking that Wine would use `read` instead of `mmap` + `memcpy` for coping the image from disk, and then we wouldn't see it here. Unless I'm misunderstanding?

samply/src/linux_shared/mod.rs
```diff
@@ -399,7 +417,54 @@ where
...
+    ///
+    /// When this misalignment happens, most of the sections in the memory will not be a file
+    /// mapping. However, the PE header is always mapped, and it resides at the beginning of the
+    /// file, which means it's also always *aligned*. Hence, if we find a mapping that seems to
+    /// pointing to a PE file, and has a file offset of 0, we'll add it to the list of "suspected
+    /// PE images". When we see a later mapping that belongs to one of the suspected PE ranges,
+    /// we'll match the mapping with the file, which allows binary correlation and unwinding to
+    /// work.
+    fn check_for_pe_mapping(&mut self, path_slice: &[u8], base_avma: u64) {
```
mstange, line 435: Let's call the `base_avma` parameter `mapping_start_avma` instead - at this point we don't know yet whether it's actually the base AVMA of an image.

samply/src/linux_shared/mod.rs
mstange: Oh I see, it mmaps just the header. Yes, the comment addition you're proposing sounds perfect.

ishitatsuyuki: I added the table and also clarified that PE header will be mapped first.

mstange:
Sorry about the Linux clippy failures. Could you run `cargo clippy --fix && cargo fmt` on Linux as the first or last commit? Then I can merge the PR.

mstange (approved):
This is really good. Thank you!

The lack of test coverage is a bit of a problem, but it's an existing problem. If you have ideas for how to improve it, I'd be happy to hear them. Maybe we can make a tool that allows creating minimal testcases from real perf.data files, by stripping out event records.

ishitatsuyuki:
Thank you for your review!

I agree that there should be more automated testing of this. Using postprocessed perf.data file sounds like a good approach to me too. We could feed in the startup and mmap events and check that the module list is as expected.

PR #29, opened 2023-02-27
Migrate to 2021 edition.

ishitatsuyuki:
Imports for TryInto and TryFrom are removed now that they are in prelude, otherwise no code changes.

This is for https://doc.rust-lang.org/edition-guide/rust-2021/disjoint-capture-in-closures.html which I'm planning for using in the sorter for #26.


mstange:
Excellent!

PR #30, opened 2023-02-27
perf_event: Remove the size limit from EventIter.

ishitatsuyuki:
To sort perf events on the fly, all available events from one stream must be examined at once, which is not possible with the current EventIter restricted to up to 31 event at once.

Rewrite the out-of-order commit logic so that pending commits are tracked in a binary heap, which will remain small as long as events are dropped in-order but can also comfortably accommodate an unbounded amount of out-of-order drop.

This changes the event iteration order. Previously, we iterated over 31 events from each buffer before going on to the next. Now, all available events in each buffer will be consumed before going on to the next. This incurs more buffering than before, but live event sorting will incur even more buffering, so seeing this as a transition step it should be acceptable.

### DIFF
--- a/samply/src/linux/perf_event.rs
+++ b/samply/src/linux/perf_event.rs
@@ -1,5 +1,5 @@
 use std::cmp::max;
-use std::fmt;
+use std::collections::BinaryHeap;
 use std::io;
 use std::mem;
 use std::ops::Range;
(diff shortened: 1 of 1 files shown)


mstange (approved):
Looks good to me! I hadn't fully understood the old code or what it was doing with the bitmask, but the new behavior is much easier to understand in my opinion.

Thank you for this great work.

PR #31, opened 2023-03-06
docs: Add alternatives for Linux setup using capabilities

jjerphan:
Hi,

First, thank you for developing samply! :handshake:

I think the setup on Linux can be adapted so that one can use [capabilities](https://man7.org/linux/man-pages/man7/capabilities.7.html) instead of removing all checks by modifying `kernel.perf_event_paranoid`.

Also, there was a typo, and according to [this piece of documentation in the source of the kernel](https://www.kernel.org/doc/Documentation/sysctl/kernel.txt), `kernel.perf_event_paranoid` must be set to `-1` so that all users can inspect events.

What do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,11 +30,25 @@ All data is kept locally (on disk and in RAM) until you choose to upload your pr
 
 samply is a sampling profiler and collects stack traces, per thread, at some sampling interval (the default 1000Hz, i.e. 1ms). On macOS, both on- and off-cpu samples are collected (so you can see under which stack you were blocking on a lock, for example). On Linux, only on-cpu samples are collected at the moment.
 
-On Linux, as samply needs access to performance events system by unprivileged users, run:
+On Linux, samply needs access to performance events system for unprivileged users. For this, you can either:
 
-```
-sudo sysctl kernel.perf_event_paranoid=1 
-```
+ - if using Linux 5.8 or later, set the `CAP_PERFMON` capability as effective and permitted for samply (recommended):
+
+   ```
+   sudo setcap 'cap_perfmon+ep' `which samply`
+   ```
+
+ - set the `CAP_SYS_ADMIN` capability as effective and permitted for samply:
+
+   ```
+   sudo setcap 'cap_sys_admin+ep' `which samply`
+   ```
+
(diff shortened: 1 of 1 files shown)


ishitatsuyuki:
`perf_event_paranoid = 1` works fine for normal profiling, where interrupts are generated every X cycle. For access to other PMUs you might need `perf_event_paranoid = 0`. I think even 3 is fine if you don't need kernel callchains, but right now that configuration requires tweaking samply source directly.

There's a known issue where mmap will fail due to mlock limit (#25), which could be mitigated by `perf_event_paranoid = -1`, but I don't think there's any issues with "all users inspecting events", provided the process is running under the same UID.

jjerphan:
> I think even 3 is fine if you don't need kernel callchains

Yep, `perf_event_paranoid = -1` allows inspecting kernel events. Should I make it more precise?

ishitatsuyuki:
It's about kernel callchains, which is just the kernel part of callstack when sampling occurs. This is available at paranoid = 1. At <=0 you can additionally access raw tracepoint but that's not used by samply right now.

I think the description doesn't need to be too precise, the main purpose is to guide users to set up their system so that samply won't fail on `perf_event_open`. Up to you, though.

One feedback I have would be that assigning `CAP_SYS_ADMIN` is more dangerous than lowering perf_event_paranoid, as the former is essentially setuid/sudo on the executable. Can we keep the suggested options to only CAP_PERFMON and perf_event_paranoid?

README.md
```diff
@@ -30,11 +30,25 @@ All data is kept locally (on disk and in RAM) until you choose to upload your pr
...
+   ```
+   sudo setcap 'cap_perfmon+ep' `which samply`
+   ```
+
+ - set the `CAP_SYS_ADMIN` capability as effective and permitted for samply:
+
+   ```
+   sudo setcap 'cap_sys_admin+ep' `which samply`
+   ```
+
```
jjerphan, line 46:
```suggestion
```

mstange:
Thanks @jjerphan for the contribution and @ishitatsuyuki for the feedback. I don't have a lot of knowledge in this area so I'll just trust the two of you and merge this. If people disagree with this advice, they can file issues.

PR #34, opened 2023-03-16
Add support for gathering perf map information in `load`

bnjbvr:
This adds support for reading information of `perf` mappings that can be generated by runtimes, using a [very simple format](https://github.com/torvalds/linux/blob/master/tools/perf/Documentation/jit-interface.txt). It's implemented only for the `load` command, until a nice design is found for the `record` step.

Unfortunately, while this correctly loads the symbols from the mapping on my machine, then when the report is displayed in Firefox Profiler's UI, the JIT symbols don't show up. Am I missing something here?


mstange:
I've rebased this against the Jitdump work and added a commit on top which I think should make it work. I haven't tested it yet, though. I'll merge it anyway, and then you can test if it works.

mstange (approved):
Thank you for the contribution!

samply/src/linux_shared/mod.rs
```diff
@@ -252,6 +255,111 @@ where
         }
     }
 
+    fn process_perf_map_line(line: &str) -> Option<(u64, u64, &str)> {
+        let mut split = line.splitn(3, ' ');
+        let addr = split.next()?;
+        let len = split.next()?;
+        let symbol_name = split.next()?;
```
mstange, line 262: I've changed this to use `splitn` so that we wouldn't need to glue together the last piece with spaces again. I hope perf.map files always use just a single space as a separator.

mstange: Oh, yeah, [the "spec"](https://github.com/torvalds/linux/blob/master/tools/perf/Documentation/jit-interface.txt) says:

> Each line has the following format, fields separated with spaces:

bnjbvr:
Works perfectly now, with symbols displayed in the Firefox Profiler UI, thanks a bunch :heart:

samply/src/linux_shared/mod.rs
bnjbvr: Sweet, TIL `splitn`.

PR #36, opened 2023-03-21
Move perf-map loading into the `Process`

bnjbvr:
So I ran into a fun bug: the process I'm profiling has a self-update mechanism, so it would first start, find out there's an update, download it and restart itself (with the same PID). The perf map was loaded once for the first process; the whole `Process` is removed with the perf event indicating the execve event. Then the `Process` is recreated, but the perf map isn't reloaded, because we incorrectly marked it as already-loaded before.

This moves the code around so it's now the responsibility of the `Process` itself to keep track if it's already loaded the perf map or not, and we do check it once per lifetime of `Process`. This fixes the bug I was observing, and I can now see the perf maps correctly displayed in the Profiler UI :partying_face:


mstange:
Thanks!

PR #38, opened 2023-03-25
Use the path of the original file to find the debuglink destination

### LINKED ISSUE #37: Wrong symbols for Rust binary
When profiling a Rust binary, it seems the symbols are not correct. Instead of the function names for all my binary I get `fun_XXXX` where `XXXX` is an hexadecimal number. However function names for C libraries we depend on are properly displayed.

I know that the symbols are available because using e.g. [flamegraph](https://github.com/flamegraph-rs/flamegraph), the proper function names show up.

I would like to help debug this but the only error I'm seeing is this in the web console:

```
17:38:33,589 SymbolsNotFoundError: Could not obtain symbols for [vdso]/000000000000000000000000000000000.
 - Error: The symbol server does not have symbols for [vdso]/000000000000000000000000000000000.
 - Error: No connection to the browser, cannot run querySymbolicationApi
 - Error: No connection to the browser, cannot obtain symbol tables
    Fi errors.js:16
    _getSymbolTablesFromBrowser symbol-store.js:504
symbolication.js:915:14
```

zecakeh:
As defined in https://www-zeuthen.desy.de/unix/unixguide/infohtml/gdb/Separate-Debug-Files.html:

> For the “debug link” method, gdb looks up the named file in the directory of the executable file, then in a subdirectory of that directory named .debug, and finally under the global debug directory, in a subdirectory whose name is identical to the leading directories of the executable's absolute file name. 

The paths were hardcoded by using the directories provided in the examples, which might not work, especially if the app is not installed via a package.

I'm not 100% sure of the assumptions I made to get the original file location as a `Path`.

Note that in the three cases it uses an absolute path but it might be enough to use a relative path for the two first locations.

Fixes #37 for me.

### DIFF
--- a/samply-symbols/src/elf.rs
+++ b/samply-symbols/src/elf.rs
@@ -65,7 +65,7 @@ where
     let debug_id = debug_id_for_object(elf_file)?;
     let name = std::str::from_utf8(name).ok()?;
     let candidate_paths = helper
-        .get_candidate_paths_for_gnu_debug_link_dest(name)
+        .get_candidate_paths_for_gnu_debug_link_dest(original_file_location, name)
         .ok()?;
 
     for candidate_path in candidate_paths {
--- a/samply-symbols/src/shared.rs
+++ b/samply-symbols/src/shared.rs
@@ -320,6 +320,7 @@ pub trait FileAndPathHelper<'h> {
     /// TODO
     fn get_candidate_paths_for_gnu_debug_link_dest(
         &self,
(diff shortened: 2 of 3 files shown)


mstange (approved):
Looks good, thank you for fixing this!

mstange:
> I'm not 100% sure of the assumptions I made to get the original file location as a `Path`.
> 
> Note that in the three cases it uses an absolute path but it might be enough to use a relative path for the two first locations.

I think what you have is fine.

PR #48, opened 2023-05-15
Fix the `mainThreadIndex` values for counters during serialization 

canova:
Previously we were using the sorted process indices for `first_thread_index_per_process` array. But when we are reading values from  that array, we are using process handles and not the sorted array indices.

Because of this we were putting the wrong `mainThreadIndex` for counters. With this change, we are now using the process handles everywhere.


mstange (approved):
Thank you!

PR #50, opened 2023-06-19
Replace webbrowser with opener.

ishitatsuyuki:
This allows launching browser correctly even within container runtimes such as the Steam Runtime or Flatpak.

This also allows the user to override the auto-launched browser via $BROWSER, in case their default browser is incompatible with Firefox Profiler. See the linked issue.

Closes: https://github.com/mstange/samply/issues/35


ishitatsuyuki:
(On the topic of other changes I currently in my fork, I want to eventually upstream PE support as well as perf event reordering, but haven't got enough time to tidy up the patches. So here's a few small QoL fixes for now.)

mstange:
Thank you!

PR #52, opened 2023-07-07
Gate `query_json_api` behind a feature flag.

afranchuk:
This adds an `api` flag to possibly reduce the dependency chain (e.g. those huge `yaxpeax-*` crates in `samply-api`).

I considered also adding a `network` flag to disable `reqwest` and `symsrv`, however the line dividing code using these crates isn't as clear (some debuginfod and other things would have to be removed or replaced with panics). I might do this in the future if vendoring/auditing becomes a chore (not to mention it's a nice way to ensure that no network access will occur).

### DIFF
--- a/wholesym/Cargo.toml
+++ b/wholesym/Cargo.toml
@@ -11,9 +11,15 @@ repository = "https://github.com/mstange/samply"
 license = "MIT OR Apache-2.0"
 keywords = ["symbols", "profiling", "addr2line", "debuginfo"]
 
+[features]
+default = ["api"]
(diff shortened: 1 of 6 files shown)


wholesym/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "wholesym"
-version = "0.3.0"
+version = "0.3.1"
```
mstange, line 3: Please keep the version unchanged, I will bump it when I publish.

mstange (approved):
Awesome, thank you!

PR #54, opened 2023-07-28
Display the absolute address for relative jumps on x86

jrmuizel:
yaxpeax doesn't have support for doing this natively so just shim it on top.


mstange:
Excellent, thanks!

PR #61, opened 2023-08-24
Linux: Use mio to get ring buffer notifications. 

ishitatsuyuki:
This makes the polling happen with epoll() instead of poll(), to cut down on FD registration overhead (there can be lots of perf FDs when attaching to a multithreaded process). It's also edge triggered, reducing false-positive notifications.

The reason to use mio instead of native epoll is:
- mio is a safe abstraction which saves us from inventing another.
- We already depend on mio through other async web runtime stuff.
- It integrates with native FDs without much boilerplate.

This also includes a version bump in lockfile for `proc-macro2` as it was breaking nightly build.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -167,12 +167,11 @@ dependencies = [
 
 [[package]]
 name = "bstr"
-version = "1.5.0"
+version = "1.6.0"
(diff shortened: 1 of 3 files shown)


mstange (approved):
Cool!

PR #62, opened 2023-08-24
Linux: Implement perf live event reordering.

ishitatsuyuki:
The perf ring buffers are divided by CPUs and processes and naively concatenating read events can yield wrong ordering. In particular, this is problematic for fork and exec events since mmaped modules will go into the wrong process and cause unwind to fail.

Introduce a streaming event sorter to tackle this. Note that timestamps should not be considered 100% reliable and cases where timestamps go backwards should be still gracefully handled. In particular for perf, a NMI could interrupt event writes, thus even timestamp inversion within the same ring buffer can happen.

This is based on top of #61 for convenience, but still split into commits. I can rebase on top of main if needed.

Link: https://github.com/mstange/samply/issues/26

### DIFF
--- a/samply/src/linux/mod.rs
+++ b/samply/src/linux/mod.rs
@@ -3,4 +3,5 @@ mod perf_group;
 mod proc_maps;
 mod process;
 pub mod profiler;
+mod sorter;
 mod sys;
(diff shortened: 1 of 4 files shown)


mstange (approved):
Looks great, thank you so much!!

Since it's not used by the perf file importer at the moment, I would move it to `samply/src/linux` rather than `linux_shared` - this will fix the macOS build warnings. Were you planning to use it for the perf file importer in the future?

ishitatsuyuki:
> Were you planning to use it for the perf file importer in the future?

It makes sense, but I have no problem with the current file importer either. I have moved it to the `linux` module.

PR #65, opened 2023-08-27
Decouple `serde` from its `derive` crate

CryZe:
By not activating the `derive` feature on `serde`, the compilation speed can be improved by a lot. This is because `serde` can then compile in parallel to `serde_derive`, allowing it to finish compilation possibly even before `serde_derive`, unblocking all the crates waiting for `serde` to start compiling much sooner.

As it turns out the main deciding factor for how long the compile time of a project is, is primarly determined by the depth of dependencies rather than the width. In other words, a crate's compile times aren't affected by how many crates it depends on, but rather by the longest chain of dependencies that it needs to wait on. In many cases `serde` is part of that long chain, as it is part of a long chain if the `derive` feature is active:

`proc-macro2` compile build script > `proc-macro2` run build script > `proc-macro2` > `quote` > `syn` > `serde_derive` > `serde` > `serde_json` (or any crate that depends on serde)

By decoupling it from `serde_derive`, the chain is shortened and compile times get much better.

Check this issue for a deeper elaboration:
https://github.com/serde-rs/serde/issues/2584

For `samply` I'm seeing a reduction from 5.43s to 3.70s when compiling `fxprof-processed-profile` in `release` mode.

image

The impact is even larger for crates that depend on `fxprof-processed-profile` such as `wasmtime`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -644,6 +644,7 @@ dependencies = [
  "debugid",
  "fxhash",
  "serde",
+ "serde_derive",
  "serde_json",
 ]
 
@@ -654,6 +655,7 @@ dependencies = [
  "assert-json-diff",
  "debugid",
  "serde",
+ "serde_derive",
  "serde_json",
 ]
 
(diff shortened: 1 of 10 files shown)


mstange (approved):
That's a great result! Thank you!

PR #68, opened 2023-08-28
Add --profile-name to record command

rukai:
Adds a `--profile-name` option to the record command that sets the "product" field in the output json so that the name in the top left hand corner of the profiler UI is set to the provided value.

I run a script that profiles my application under various scenarios and then opens all the results in my browser.
Currently the name in the corner is just set to the pid of the process which does not allow me to distinguish which tab corresponds to which scenario.
This PR fixes that.

Hovering over the profile name in the UI yields a tooltip "Edit the profile name".
So I think `--profile-name` is definitely the best name to expose to users

Example:
image

Marked as draft for now to see if mac os passes CI.

### DIFF
--- a/samply/src/linux/profiler.rs
+++ b/samply/src/linux/profiler.rs
@@ -32,6 +32,7 @@ pub type ConvertRegsNative = crate::linux_shared::ConvertRegsAarch64;
 #[allow(clippy::too_many_arguments)]
 pub fn start_recording(
     output_file: &Path,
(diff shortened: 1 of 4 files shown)


mstange:
This is an excellent idea.

rukai:
I fixed the one mac os compiler error I got on CI, hopefully there wasnt more hiding behind the next stage of compilation.

PR #70, opened 2023-09-04
Switch to SwCpuClock on ENOENT error

rkd-msw:
The `perf_event_open` manpage (https://www.man7.org/linux/man-pages/man2/perf_event_open.2.html) says that ENOENT is "Returned if the type setting is not valid.". Prior to this fix, samply only switched to SwCpuClick on an EPERM error. This fix means that we try it speculatively whatever the error, unless it's a EPERM that can be fixed by changing the paranoia level.

(I've seen an ENOENT error when running samply on WSL2, where hardware counters aren't available.)

### DIFF
--- a/samply/src/linux/profiler.rs
+++ b/samply/src/linux/profiler.rs
@@ -353,11 +353,10 @@ fn init_profiler(
         attach_mode,
     );
 
-    let mut perf = match perf {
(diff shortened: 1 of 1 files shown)


mstange:
Thanks!

PR #72, opened 2023-09-17
use libc::SYS_perf_event_open to linux sycall

Tudyx:
This allows:  

- not redefine it our-self
- support all the architecture used by libc


mstange:
Thanks!

PR #76, opened 2023-11-07
Enable samply to run on Android

flxo:
Android is pretty much `linux` but needs to be enabled in the `target_os` guards. Tested on a headless Android 9 with 5.4.


mstange:
Thanks! I'll fix the clippy failures on main and then you can rebase.

mstange:
Can you check how hard it would be to add CI for this configuration? Without CI it's likely to break in the future.

flxo:
Thanks. Regarding CI: sure - I've had the combination GH actions + target aarch64-ilnux-android a couple of times. I'll prepare a PR.

PR #79, opened 2023-11-14
Extend Rust CI setup

flxo:
Revise the CI matrix to adopt a target-centric approach. Clearly define                                                                                                                                                                               
the operating system for each target, outlining the specific OS on which        
the build steps are intended to execute.        
        
Ensure the installation of cross where necessary for individual targets.        
Note that for the aarch64-linux-android target, a custom revision is        
currently required instead of relying on the latest release for proper        
linking. It is anticipated that this requirement will be addressed in        
the upcoming release of cross (refer to: https://github.com/cross-rs/cross/issues/135).        
        
Add targets `aarch64-linux-android`, `aarch64-unknown-linux-gnu` and        
`x86_64-unknown-linux-musl`.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -10,17 +10,63 @@ jobs:
 
     strategy:
       matrix:
-        os: [ubuntu-latest, macos-latest, windows-latest]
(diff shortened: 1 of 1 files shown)


mstange:
Excellent, thanks!

PR #85, opened 2024-01-18
Update framehop to 0.8.0.

afranchuk:


afranchuk:
:facepalm: didn't update the mac code.

afranchuk:
Looks like the mac code works now, though I am unable to test it.

samply/src/linux_shared/converter.rs
```diff
@@ -1088,24 +1075,28 @@ where
...
                 base_svma,
-                text: text.as_ref().map(svma_range),
-                text_env: text_env.as_ref().map(svma_range),
-                stubs: None,
-                stub_helper: None,
-                eh_frame: eh_frame.as_ref().map(svma_range),
-                eh_frame_hdr: eh_frame_hdr.as_ref().map(svma_range),
-                got: got.as_ref().map(svma_range),
+                text_svma: text.as_ref().map(svma_range),
+                text: text_data,
```
mstange, line 1081: There is a bug here. The text data is no longer associated with an address range. I think it is now assumed to cover the text **section** range. However, that's not the case here: text_data is covering the `__TEXT` **segment** range. The `__TEXT` segment starts at the base address and includes the `__text` section as well as other executable sections, such as the `__stubs` and `__stub_helper` sections.

This is causing crashes.

mstange: Er, this diagnosis is nonsense. I am commenting on the Linux code here. I will look into it in more detail.

afranchuk: FWIW I was suspicious of the macos code as I couldn't test it easily right now. Right now it passes the __TEXT segment in as the `text` section, only because (and this isn't a good reason) that's what was previously done. https://github.com/mstange/samply/pull/85/files#diff-a0f24fc85b21b64ae90929d488493b6409e4afabf372e030bc93440b517cc1dfR481

Switching that out for the __text section will probably fix the issue (dunno why I thought I should copy what it was doing before).

mstange: The crash was happening because `VmData` has a garbage `Clone` implementation and this PR was adding the first actual use of clone.

afranchuk: FWIW the clone was only needed to clone that `__TEXT` I think, so changing that code might have fixed it too!

mstange:
Merged as https://github.com/mstange/samply/commit/f387ef0ecf461e1e67577b3c7fb57d5dc2fbc01d with some changes to the macOS part.

Thanks!

PR #88, opened 2024-02-10
Add dependabot

oriontvv:
This PR adds dependabot config

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,15 @@
+version: 2
+updates:
+  - package-ecosystem: "cargo"
+    directory: "/"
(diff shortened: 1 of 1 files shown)


mstange (changes requested):
Thanks, I'd be happy to accept this if you update it to match the changes that have been made to https://github.com/actix/actix-web/blob/master/.github/dependabot.yml :

 - We don't have a use for the "docker" part, as far as I can tell.
 - The quotes are unnecessary
 - I'm fine with a weekly cadence

oriontvv:
Done

mstange (approved):
Thanks!

PR #110, opened 2024-03-08
Add categories for markers in MarkerTable.

yjugl:
* Introduce a new MarkerTableEntry type to alter a marker table entry after creation. It can add a category for the marker and will be easy to extend while preserving the public API in the future.

* Change the return type of `add_marker()` and `add_marker_with_stack()` functions from `()` to `MarkerTableEntry`. This means the `add_*` calls can now be chained with methods that apply on a MarkerTableEntry to alter the entry that was created; in particular to add a category.

This patch changes the public API of the crate. If users of the crate were calling `add_*` methods without a trailing `;`, they will now need to add that trailing `;`.

### DIFF
--- a/fxprof-processed-profile/src/marker_table.rs
+++ b/fxprof-processed-profile/src/marker_table.rs
@@ -1,14 +1,13 @@
 use serde::ser::{Serialize, SerializeMap, Serializer};
 use serde_json::Value;
 
-use crate::serialization_helpers::{
-    SerializableOptionalTimestampColumn, SerializableSingleValueColumn,
-};
+use crate::serialization_helpers::SerializableOptionalTimestampColumn;
 use crate::thread_string_table::ThreadInternalStringIndex;
-use crate::{MarkerTiming, Timestamp};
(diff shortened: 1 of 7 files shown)


yjugl:
Example of how to use the new API [here](https://github.com/yjugl/etw-profiler/blob/marker-categories/etw-gecko/src/main.rs#L1115): `profile.add_marker(thread.handle, name, TextMarker(text), timing).category(category);`.

mstange:
It's a neat trick, but at this stage I'm not too concerned about breaking compatibility. I think it's fine to make the category a required parameter. It looks a bit odd to have just one optional field. We can reconsider this in the future if we want to add any other optional parameters.

yjugl:
> It's a neat trick, but at this stage I'm not too concerned about breaking compatibility. I think it's fine to make the category a required parameter. It looks a bit odd to have just one optional field. We can reconsider this in the future if we want to add any other optional parameters.

Alright, I've updated the pull request to just have the category as an argument.

PR #128, opened 2024-03-27
Tag samples with time as close as possible to the time they were taken

vvuk:
This is a fix for #127 -- use the monotonic time for a sample as close as possible to the time the sample was taken. This seems sufficient to fix the issue; I'm not sure what the impact of the `Timestamp` being off is going to be. I _think_ that's safe, because that timestamp is (again, I think) what's used to group/line up the samples across threads/tasks. But the monotonic time is used for lib op resolution.


mstange:
Thanks!

PR #136, opened 2024-04-08
feat: add native allocations table and memory sample methods

Rjected:
In using this repo for converting jemalloc profiles to firefox profiler profiles, I needed to add the following:
 * Introduces the `NativeAllocations` table to the firefox processed profiles.
 * Adds copies of the existing documentation for `SampleTable` and `StackTable` from `firefox-devtools/profiler`
 * Introduces `add_memory_sample` for the `Profile` and `Thread` structs, to populate the native allocations table.

### DIFF
--- a/fxprof-processed-profile/src/profile.rs
+++ b/fxprof-processed-profile/src/profile.rs
@@ -427,6 +427,36 @@ impl Profile {
         self.threads[thread.0].add_sample(timestamp, stack_index, cpu_delta, weight);
     }
 
+    /// Add a sample to the given thread.
+    ///
+    /// The sample has a timestamp, a stack, a CPU delta, and a weight.
+    ///
+    /// The stack frames are supplied as an iterator. Every frame has an associated
+    /// category pair.
+    ///
+    /// You can can also set the weight to something negative, such as -1, to create a
+    /// "diff profile". For example, if you have partitioned your samples into "before"
+    /// and "after" groups, you can use -1 for all "before" samples and 1 for all "after"
+    /// samples, and the call tree will show you which stacks occur more frequently in
+    /// the "after" part of the profile, by sorting those stacks to the top.
+    pub fn add_memory_sample(
+        &mut self,
+        thread: ThreadHandle,
+        timestamp: Timestamp,
+        frames: impl Iterator<Item = FrameInfo>,
+        memory_address: Option<usize>,
+        weight: i32,
+    ) {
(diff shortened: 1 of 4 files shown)


mstange (approved):
Thank you! I think I'll rename some things and tweak some terminology in the comments, but this looks good!

For example, I think `add_allocation_sample` would be a better name. The name `add_memory_sample` makes me think that we are recording a single number which is the current memory usage.

Rjected:
> Thank you! I think I'll rename some things and tweak some terminology in the comments, but this looks good!
> 
> For example, I think `add_allocation_sample` would be a better name. The name `add_memory_sample` makes me think that we are recording a single number which is the current memory usage.

right, that makes sense!

mstange:
I've made some tweaks in https://github.com/mstange/samply/commit/b9d9cb3df409c93bad826ac49151c1d894437b84 .

This makes the allocation address mandatory. Does that work for your use case?

Rjected:
> I've made some tweaks in [b9d9cb3](https://github.com/mstange/samply/commit/b9d9cb3df409c93bad826ac49151c1d894437b84) .
> 
> This makes the allocation address mandatory. Does that work for your use case?

sorry, just saw this - the jemalloc profiles don't have the allocation address, so I think this actually does not work with my use case

a profile format example is here:
https://jemalloc.net/jemalloc.3.html#heap_profile_format

Rjected:
Reading the commit message:
>  - Make the memory address mandatory. The Firefox Profiler expects this
   column to either be present completely (with non-null values for every
   sample) or absent completely - and if absent, the threadId column must
   be absent too. So we make it present completely.

Jemalloc profiles don't have memory addresses afaict, but do have thread nums. However, I haven't implemented parsing per thread, so an API requiring either both or none should work

mstange:
Ah, then maybe we should add another method called `add_basic_allocation_sample` and do something (panic? no-op?) if the two are mixed.
It's too bad that you don't get the allocation address. Emitting an address per allocation is easier than emitting a stack. I did some brief digging through the jemalloc source, and I think both pieces of information are present around here: https://github.com/jemalloc/jemalloc/blob/fa451de17fff73cc03c31ec8cd817d62927d1ff9/src/jemalloc.c#L2567-L2582

I think the stack is sampled during `prof_alloc_prep`. And `allocation` is the pointer to the allocated object, and then gets passed to `prof_malloc`.

Without the address, you're not going to get a "Retained memory" view in the profiler, and that view can be quite useful.

PR #140, opened 2024-04-17
Ensure new non-main threads have their names set

jamienicol:


jamienicol:
Before: https://share.firefox.dev/3Ui2gF8
After: https://share.firefox.dev/4aSsmny

Ignore the "Thread-2", "Thread-3" etc in the after profile. Those are indeed the "correct" names due to https://bugzilla.mozilla.org/show_bug.cgi?id=1891726

mstange:
Oh yeah, this looks like an oversight.

I think this wasn't needed on Linux for the following reasons:
- For new threads, we first inherit the name from the parent (while handling `FORK` records) and then update the name once we see the `COMM` record.
- For existing threads when `samply record` does the profiling, we call `register_existing_thread` with the name from `/proc/{pid}/task/{tid}/comm`.
- For existing threads when `perf record` does the profiling, [perf synthesizes `COMM` events at the start of the trace](https://elixir.bootlin.com/linux/v6.8.7/source/tools/perf/util/synthetic-events.c#L819).

So it seems either simpleperf does not synthesize COMM events, or maybe it puts the "thread map" in a different place.

Anyway, this change looks like a clear improvement, merging.

PR #142, opened 2024-04-20
Add --unlink-aux-files option

### LINKED ISSUE #64: Consider unlinking jitdump file after open
Cleanup of jitdump files is a bit of a pain. It would be handy if samply just unlinked them after it opens them -- at that point both the generating process and samply should have an open fd to the file. Could be done only with a flag in case the files are useful for something else (or maybe a flag that's `--no-unlink-jitdump` or something, with unlink being the default).

vvuk:
Adds an option to `record` to unlink jitdump/marker files after they've been read. Fixes #64.

.. though looking over this, I only added it to mac, will add to linux.

### DIFF
--- a/samply/src/main.rs
+++ b/samply/src/main.rs
@@ -180,6 +180,12 @@ pub struct ProfileCreationArgs {
     /// Fold repeated frames at the base of the stack.
     #[arg(long)]
     fold_recursive_prefix: bool,
+
+    /// If a process produces jitdump or marker files, unlink them after
+    /// opening. This ensures that the files will not be left in /tmp,
+    /// but it will also be impossible to look at JIT disassembly.
+    #[arg(long)]
+    unlink_aux_files: bool,
 }
 
 fn main() {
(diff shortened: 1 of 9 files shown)


mstange:
This looks good, will merge once you've added Linux support and addressed the failures.

samply/src/main.rs
```diff
@@ -180,6 +180,12 @@ pub struct ProfileCreationArgs {
     /// Fold repeated frames at the base of the stack.
     #[arg(long)]
     fold_recursive_prefix: bool,
+
+    /// If a process produces jitdump or marker files, unlink them after
+    /// opening. This ensures that the files will not be left in /tmp,
+    /// but it will also be impossible to look at JIT disassembly.
```
mstange, line 186: `[...], and line number information will be missing in the profiles.`

vvuk:
Actually, I just had an idea -- I'll make this flag unlink jitdump files after the samply process exits. This would mean that while the server is up, the jitdump files are still around to provide all the info, but get cleaned up at exit. And if you're doing --save-only, well, don't pass this flag.

mstange:
Well, that'll work until I make the planned change to daemonize the server, so that you don't have to keep pressing Ctrl+C after every profiling run (and so that you can serve multiple profiles at the same time without using up excessive numbers of ports).

I'd say don't worry about it too much at the moment - do whatever's easy and solves your use case, and we can sort out a long term solution later, once I've put the cards on the table by filing issues with all the things I have in mind.

PR #170, opened 2024-05-04
Pull etw-reader back in at the root

vvuk:


mstange:
Thanks a ton!

Merging into etw-reader-at-root branch now so that I can add commits.

PR #175, opened 2024-05-06
Only request elevation once

mstange:
The UAC prompt when stopping the profiler is very annoying.

This PR adds a helper process that only needs to be elevated once and which launches both xperf processes. In the future we'll be able to use this helper process to implement more of what xperf is doing at the moment.

I got a bit carried away during the implementation. There's probably a simpler implementation hiding in there. The current implementation uses no less than four lock files, and two more files for the actual IPC messages, all stored in a TempDir.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1842,6 +1842,7 @@ dependencies = [
  "etw-reader",
  "flate2",
  "framehop",
+ "fs4",
  "futures-util",
  "fxhash",
  "fxprof-processed-profile",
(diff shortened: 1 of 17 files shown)


mstange:
> I got a bit carried away during the implementation. There's probably a simpler implementation hiding in there. The current implementation uses no less than four lock files, and two more files for the actual IPC messages, all stored in a TempDir.

I decided to stick with the somewhat over-engineered implementation. It's not just the helper process that needs to block on receiving the "stop" message from the parent; the parent also needs to wait for the helper process to tell it when xperf has started, so that it doesn't launch the recorded process too early. And any errors from the helper process need to be propagated to the parent.

PR #186, opened 2024-05-08
Windows ARM64 fixes, some command line arg stuff

vvuk:
Fixes up arm64 cpu deltas, and adds vm_hack/coreclr arguments to record for Windows.


mstange (approved):
Looks good!

PR #189, opened 2024-05-11
Handle CoreCLR events

vvuk:
This adds support for enabling the CoreCLR ETW provider and tracking events related to
JIT compilation, GC, and others if the `--coreclr` argument is used. There is also
a `--coreclr-allocs`, which will emit an event for every GC allocation (which is a lot).
(The allocation tracking doesn't emit the type name yet, just a type handle, but names
are coming.)

CoreCLR events are only currently useful if Samply launches the app; there's future work
coming to enable the CoreCLR "rundown" provider that provides all the necessary info
at the start of capture for existing processes. (Similar to the DCStart events; it's just
not enabled by default because it's very expensive, as it has to dump info about all JIT
methods already compiled etc.)


mstange (approved):

mstange:
> CoreCLR events are only currently useful if Samply launches the app;

Or when using `samply record -a` if the app starts after profiling starts, I assume.

PR #190, opened 2024-05-11
Allow attaching to existing macOS process

vvuk:
Implement attaching to existing mac processes. This needs samply to be signed with the debugger entitlement:

1. Create a file `ent.xml`:
```
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
<dict>
	<key>com.apple.security.cs.debugger</key>
	<true/>
</dict>
</plist>
```
2. `cargo build`
3. `codesign --force --options runtime --sign - --entitlements ent.xml target/debug/samply`

The basic functionality works, but there is still work to do:

1. This doesn't capture jitdump or markers. It's possible to just watch /tmp/ for appropriately-named jitdump files, but a real solution would need cooperation with the existing process to notify it to dump all existing jit compiled info to jitdump files.
2. This won't capture child processes. That relies on the dyld preload to send the task of every new process back, and there's no preload here because there's no env vars set.

For 2, this might be possible to sort out if can set the dyld preload env var in the target process. I don't know if there's a simpler way to do this, but one idea is to load the preload shared library into the target (I think this is doable?) and then create a new thread in that process that calls an init function from the preload lib which will set the process env.

### DIFF
--- a/samply/src/mac/process_launcher.rs
+++ b/samply/src/mac/process_launcher.rs
@@ -238,10 +238,7 @@ pub struct AcceptedTask {
 }
 
 impl AcceptedTask {
-    pub fn take_task(&mut self) -> mach_port_t {
-        // TODO: I think it's safe to not do this mem::replace; we may need to resume the task
-        // in start_execution()
-        //mem::replace(&mut self.task, MACH_PORT_NULL)
+    pub fn task(&self) -> mach_port_t {
         self.task
     }
 
--- a/samply/src/mac/process_launcher.rs
+++ b/samply/src/mac/process_launcher.rs
@@ -5,24 +5,97 @@ use std::io::Write;
 use std::mem;
 use std::os::unix::prelude::OsStrExt;
 use std::path::PathBuf;
-use std::process::{Child, Command};
+use std::process::{Child, Command, ExitStatus};
 use std::sync::Arc;
 use std::time::Duration;
 
 use flate2::write::GzDecoder;
+use mach::task::{task_resume, task_suspend};
+use mach::traps::task_for_pid;
 use tempfile::tempdir;
 
+use crate::shared::ctrl_c::CtrlC;
+
 pub use super::mach_ipc::{mach_port_t, MachError, OsIpcSender};
-use super::mach_ipc::{BlockingMode, OsIpcMultiShotServer, MACH_PORT_NULL};
+use super::mach_ipc::{mach_task_self, BlockingMode, OsIpcMultiShotServer, MACH_PORT_NULL};
+
+pub trait RootTaskRunner {
+    fn run_root_task(&self) -> Result<ExitStatus, MachError>;
+}
 
 pub struct TaskLauncher {
     program: OsString,
     args: Vec<OsString>,
     child_env: Vec<(OsString, OsString)>,
-    _temp_dir: Arc<tempfile::TempDir>,
+    iteration_count: u32,
+}
+
+impl RootTaskRunner for TaskLauncher {
+    fn run_root_task(&self) -> Result<ExitStatus, MachError> {
+        // Ignore Ctrl+C while the subcommand is running. The signal still reaches the process
+        // under observation while we continue to record it. (ctrl+c will send the SIGINT signal
+        // to all processes in the foreground process group).
+        let mut ctrl_c_receiver = CtrlC::observe_oneshot();
+
+        let mut root_child = self.launch_child();
+        let mut exit_status = root_child.wait().expect("couldn't wait for child");
+
+        for i in 2..=self.iteration_count {
+            if !exit_status.success() {
+                eprintln!(
+                    "Skipping remaining iterations due to non-success exit status: \"{}\"",
+                    exit_status
+                );
+                break;
+            }
+            eprintln!("Running iteration {i} of {}...", self.iteration_count);
+            let mut root_child = self.launch_child();
+            exit_status = root_child.wait().expect("couldn't wait for child");
+        }
+
+        // From now on, we want to terminate if the user presses Ctrl+C.
+        ctrl_c_receiver.close();
+
+        Ok(exit_status)
+    }
 }
 
 impl TaskLauncher {
+    pub fn new<I, S>(
+        program: S,
+        args: I,
+        iteration_count: u32,
+        env_vars: &[(OsString, OsString)],
+        extra_env_vars: &[(OsString, OsString)],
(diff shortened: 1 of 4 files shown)


vvuk:
Hrm. Being able to attach to newly created children seems very complicated, borderline not doable. In order to write to the environment, the magic seems to be the `_NSGetEnviron` symbol which returns a `char***` (pointer to the location where the array of `char*` that form the environment is stored). There are many problems with this path:

- We need to figure out the address of `_NSGetEnviron` from `libSystem`. This is maybe possible, since we have the library info for the target task, and we can load debug info... we should be able to just forward-resolve the symbol ourselves.
- Once we got the address of this symbol, in theory we should be able to read task memory of the relocated instructions and.. maybe.. compute the actual address of the environment pointer. I'm not sure if we'll have enough info.
- Assuming we get the environment pointer, we can allocate and write task memory to create a new env block
- ... but that will blow up the next time the process tries to setenv or similar [because setenv and friends use `realloc`, so the memory needs to come from `malloc`](https://opensource.apple.com/source/Libc/Libc-167/stdlib.subproj/setenv.c.auto.html)

Spawning a thread in the target and executing code seems like maybe more possible, especially if that code can be `dlopen()/dlsym()/call`. Reading through [this stuff](https://book.hacktricks.xyz/macos-hardening/macos-security-and-privilege-escalation/macos-proces-abuse/macos-ipc-inter-process-communication/macos-thread-injection-via-task-port) it pointed me to [this threadexec library](https://github.com/bazad/threadexec) for which the source is pretty damn complex. But one interesting bit is that it takes the address of a function in the local process, and uses that same address in the target process... so maybe the dyld cache is not mapped at different locations in different processes? I can't believe that wouldn't be the case for security, maybe it wasn't 6 years ago.

Given all this, I'm inclined to just not support attaching to new children when profiling a target process. It should be possible to attach to existing children, though.

mstange:
For profiling child processes, we could poll [`proc_listchildpids`](https://github.com/apple-oss-distributions/xnu/blob/94d3b452840153a99b38a3a9659680b2a006908e/libsyscall/wrappers/libproc/libproc.h#L95), if it doesn't have too much overhead. There's a Rust wrapper in [`remoteprocess::Process::child_processes`](https://github.com/benfred/remoteprocess/blob/3e94f997345565e3dfe36e1dc206abb557c13e67/src/osx/mod.rs#L138). We'd miss the very beginning of new processes, but it's probably good enough for most use cases.

mstange:
And for profiling system-wide, if it's even possible with acceptable overhead, we could list all processes using the `KERN_PROC_ALL` sysctl, like lldb does in [`Host::FindProcessesImpl`](https://github.com/llvm/llvm-project/blob/0869204cff22831d0bb19a82c99bf85e4deb4ae3/lldb/source/Host/macosx/objcxx/Host.mm#L649-L665).

mstange:
Thanks for investigating this! I was imagining using a sudo subprocess for this, but self-signing samply is a good idea too. We could even have both. And we could have a `samply setup` command that does this for the user, with some kind of interactive wizard.

samply/src/mac/process_launcher.rs
```diff
@@ -5,24 +5,97 @@ use std::io::Write;
...
+
+impl RootTaskRunner for TaskLauncher {
+    fn run_root_task(&self) -> Result<ExitStatus, MachError> {
+        // Ignore Ctrl+C while the subcommand is running. The signal still reaches the process
+        // under observation while we continue to record it. (ctrl+c will send the SIGINT signal
+        // to all processes in the foreground process group).
+        let mut ctrl_c_receiver = CtrlC::observe_oneshot();
+
+        let mut root_child = self.launch_child();
+        let mut exit_status = root_child.wait().expect("couldn't wait for child");
```
mstange, line 41: If samply has debugger entitlements, can we call `task_for_pid` for `root_child.id()` if we're launching system binaries or shell scripts? Or will those still be denied?

mstange (approved):

mstange:
> Spawning a thread in the target and executing code seems like maybe more possible, especially if that code can be `dlopen()/dlsym()/call`. Reading through [this stuff](https://book.hacktricks.xyz/macos-hardening/macos-security-and-privilege-escalation/macos-proces-abuse/macos-ipc-inter-process-communication/macos-thread-injection-via-task-port) it pointed me to [this threadexec library](https://github.com/bazad/threadexec) for which the source is pretty damn complex. 

I wanted to point you at [Listing 12-9](https://newosxbook.com/src.jl?tree=&file=/listings/inject.c) from the \*OS Internals Book 1, but then I found [this post](https://knight.sc/malware/2019/03/15/code-injection-on-macos.html) which has further improved on it, and on of the comments on [the gist with the full code](https://gist.github.com/knightsc/45edfc4903a9d2fa9f5905f60b02ce5a) links to [this implementation which is arm64 compatible](https://gist.github.com/vocaeq/fbac63d5d36bc6e1d6d99df9c92f75dc).

vvuk:
> Thanks for investigating this! I was imagining using a sudo subprocess for this, but self-signing samply is a good idea too. We could even have both. And we could have a samply setup command that does the self-signing for the user, with some kind of interactive wizard.

Yep, I was thinking `samply setup` too. Followup PR though. For root -- I think self-signing via setup and/or just calling via `sudo` directly should be sufficient, I don't think we need a separate subprocess. I swear I read somewhere that at some point Apple will disallow `task_for_pid` even for root processes if they don't have the entitlement, but I can't find it.  It works right now; same code, just running `sudo samply` without code signing.

> I wanted to point you at [Listing 12-9](https://newosxbook.com/src.jl?tree=&file=/listings/inject.c) from the *OS Internals Book 1, but then I found [this post](https://knight.sc/malware/2019/03/15/code-injection-on-macos.html) which has further improved on it, and one of the comments on [the gist with the full code](https://gist.github.com/knightsc/45edfc4903a9d2fa9f5905f60b02ce5a) links to [this implementation which is arm64 compatible](https://gist.github.com/vocaeq/fbac63d5d36bc6e1d6d99df9c92f75dc).

Awesome, thanks! Good to know, though I probably won't go down this route any time soon.

mstange:
> and/or just calling via `sudo` directly should be sufficient

I want to discourage `sudo samply record` because this would also run the webserver as root. I think it would also leave a profile.json file that's not writable without sudo.

PR #199, opened 2024-05-13
Don't error out parsing 32-bit types from 64-bit events

vvuk:
Both 32-bit and 64-bit events can contain 32 bit or 64 bit datatypes; it's only the pointer type that changes size.

### DIFF
--- a/etw-reader/src/parser.rs
+++ b/etw-reader/src/parser.rs
@@ -391,21 +391,19 @@ impl TryParse<Address> for Parser<'_> {
         let prop_info = &self.cache[indx];
 
         if let PropertyDesc::Primitive(desc) = &prop_info.property.desc {
-            if self.event.is_64bit() {
-                if desc.in_type == InTypeUInt64
-                    || desc.in_type == InTypePointer
-                    || desc.in_type == InTypeHexInt64
-                {
-                    if std::mem::size_of::<u64>() != prop_info.buffer.len() {
-                        return Err(ParserError::LengthMismatch);
-                    }
-                    return Ok(Address::Address64(u64::from_ne_bytes(
-                        prop_info.buffer.try_into()?,
-                    )));
+            if desc.in_type == InTypeUInt64
+                || desc.in_type == InTypeHexInt64
+                || (desc.in_type == InTypePointer && self.event.is_64bit())
+            {
+                if std::mem::size_of::<u64>() != prop_info.buffer.len() {
+                    return Err(ParserError::LengthMismatch);
                 }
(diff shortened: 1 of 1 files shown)


jrmuizel:
What's the event that you're running into this problem with?

vvuk:
@jrmuizel Hrm I don't have that handy -- it was a dotnet event that had a Method Address of type "pointer" inside of it. I can try to recreate it; it was coming from a 32-bit process that got caught during dotnet rundown. What's your concern? I mean  32-bit and 64-bit integer types should be supported regardless whether the event itself is flagged as 64 bit or not

vvuk:
Oh wait, I see what you're asking -- why is something trying to parse a 32-bit value as an "Address" in a 64-bit event? Does the 64-bit flag correspond to the process, or to the system/xperf trace run?

vvuk:
Ok, got a capture. CrowdStrike is our IT-mandated rookit, which is probably why it's inside the Lenovo service. The process here (`Lenovo..PluginHost.Device.exe`) is a 32-bit process.

```
event: 6239972055682 [pid: 4294967295 tid: 4294967295] MSNT_SystemTrace/Process/DCStart
  UniqueProcessKey= 0xffffa688f42430c0,   ProcessId= 12300,   ParentId= 6220,   SessionId= 1,   ExitStatus= 259,   DirectoryTableBase= 0xc02150000,   Flags= 2,   UserSID= Unknown InTypeWBEMSID -> OutTypeNull,   ImageFileName
= Lenovo.Modern.ImController.PluginHost.Device.exe,   CommandLine= -name 0f4360b0-f4cd-45ba-ab28-ca8d6ab93377 -runas -pluginName GenericMessagingPlugin -pluginVersion 3.2.0.142,   PackageFullName= ,   ApplicationId= ,
...
event: 6239972163015 [pid: 12300 tid: 12104] CLRMethodRundown/MethodDCStartVerbose
  MethodID= 0x135566c,   ModuleID= 0x1354f9c,   MethodStartAddress= 0x2fd0848,   MethodSize= 89,   MethodToken= 100663347,   MethodFlags= 8,   MethodNamespace= CrowdStrike.Sensor.ScriptControl.DllMain,   MethodName= .cctor,   MethodSignature= void  (),   ClrInstanceID= 9,
^ the above causes the crash
```

`MethodStartAddress` is explicitly a `HexInt64` in the manifest for `MethodDCStartVerbose`... I guess maybe the right solution here is for me to parse it as a u64?

vvuk:
Closing for #203 . Thanks for asking WTF :)

PR #202, opened 2024-05-15
Implement --unstable-presymbolicate

vvuk:
Adds `--unstable-presymbolicate`, which will output ` .syms.json` file alongside the regular profile output. If that file is present when the server is fired up (based on the profile name, i.e. `foo.json` -> `foo.syms.json`), it is parsed and the data is added to be used when symbolicating. If a library has presymbolicated info (by debugid), _only_ that info is used. There's no attempt to fall back to on-disk etc.

This allows a capture to be entirely self-contained, so that no symbol servers or on-disk data are needed for looking at the profile, which is very useful for running in CI or other locations where publishing the debuginfo isn't necessarily useful otherwise.

### DIFF
--- a/fxprof-processed-profile/src/global_lib_table.rs
+++ b/fxprof-processed-profile/src/global_lib_table.rs
@@ -1,3 +1,4 @@
+use std::collections::BTreeSet;
 use std::sync::Arc;
 
 use serde::ser::{Serialize, Serializer};
@@ -14,6 +15,11 @@ pub struct GlobalLibTable {
     used_libs: Vec<LibraryHandle>, // append-only for stable GlobalLibIndexes
     lib_map: FastHashMap<LibraryInfo, LibraryHandle>,
     used_lib_map: FastHashMap<LibraryHandle, GlobalLibIndex>,
+    /// We keep track of RVA addresses that exist in frames that are assigned to this
+    /// library, so that we can potentially provide symbolication info ahead of time.
+    /// This is here instead of in `LibraryInfo` because we don't want to serialize it,
+    /// and because it's currently a hack.
+    all_libs_seen_rvas: Vec<Option<BTreeSet<u32>>>,
 }
 
 impl GlobalLibTable {
@@ -23,6 +29,7 @@ impl GlobalLibTable {
             used_libs: Vec::new(),
             lib_map: FastHashMap::default(),
             used_lib_map: FastHashMap::default(),
+            all_libs_seen_rvas: Vec::new(),
         }
     }
 
@@ -31,6 +38,7 @@ impl GlobalLibTable {
         *self.lib_map.entry(lib.clone()).or_insert_with(|| {
             let handle = LibraryHandle(all_libs.len());
             all_libs.push(lib);
+            self.all_libs_seen_rvas.push(None);
             handle
         })
     }
@@ -51,6 +59,19 @@ impl GlobalLibTable {
     pub fn get_lib(&self, index: GlobalLibIndex) -> Option<&LibraryInfo> {
         let handle = self.used_libs.get(index.0)?;
         self.all_libs.get(handle.0)
+    }
+
+    pub fn add_lib_used_rva(&mut self, index: GlobalLibIndex, address: u32) {
+        let handle = self.used_libs.get(index.0).unwrap();
+        let lib_seen_rvas = self.all_libs_seen_rvas[handle.0].get_or_insert_with(BTreeSet::new);
+        lib_seen_rvas.insert(address);
+    }
+
+    pub fn lib_used_rva_iter(&self) -> UsedLibraryAddressesIterator {
+        UsedLibraryAddressesIterator {
+            next_used_lib_index: 0,
+            global_lib_table: self,
+        }
     }
 }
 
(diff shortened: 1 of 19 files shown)


fxprof-processed-profile/src/global_lib_table.rs
```diff
@@ -73,4 +94,26 @@ impl Serialize for GlobalLibIndex {
 
 /// The handle for a library, obtained from [`Profile::add_lib`](crate::Profile::add_lib).
 #[derive(Debug, Clone, Copy, PartialOrd, Ord, PartialEq, Eq, Hash)]
-pub struct LibraryHandle(usize);
+pub struct LibraryHandle(pub(crate) usize);
```
mstange, line 97: Is the `pub(crate)` still necessary?

fxprof-processed-profile/src/global_lib_table.rs
```diff
@@ -14,6 +15,11 @@ pub struct GlobalLibTable {
     used_libs: Vec<LibraryHandle>, // append-only for stable GlobalLibIndexes
     lib_map: FastHashMap<LibraryInfo, LibraryHandle>,
     used_lib_map: FastHashMap<LibraryHandle, GlobalLibIndex>,
+    /// We keep track of RVA addresses that exist in frames that are assigned to this
+    /// library, so that we can potentially provide symbolication info ahead of time.
+    /// This is here instead of in `LibraryInfo` because we don't want to serialize it,
+    /// and because it's currently a hack.
+    all_libs_seen_rvas: Vec<Option<BTreeSet<u32>>>,
```
mstange, line 22: Please add a comment saying that the Vec is indexed by `LibraryHandle.0`.

Alternatively, you could make the Vec indexed by `GlobalLibIndex.0`, needing one less translation in `add_lib_used_rva` and at that point you could also remove the Option and always initialize it to the empty set.

mstange (commented):
Will review more tomorrow.

fxprof-processed-profile/src/global_lib_table.rs
```diff
@@ -73,4 +94,26 @@ impl Serialize for GlobalLibIndex {
...
+    global_lib_table: &'a GlobalLibTable,
+}
+
+impl<'a> Iterator for UsedLibraryAddressesIterator<'a> {
+    type Item = (&'a LibraryInfo, Option<&'a BTreeSet<u32>>);
+
+    fn next(&mut self) -> Option<Self::Item> {
+        self.global_lib_table
+            .used_libs
+            .get(self.next_used_lib_index)
```
mstange, line 110: Could use a question mark here, so you don't need the `map`.

samply/src/shared/symbol_precog.rs
```diff
@@ -0,0 +1,386 @@
...
+
+impl Serialize for StringTableIndex {
+    fn serialize<S: Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
+        self.0.serialize(serializer)
+    }
+}
+
+impl<'de> Deserialize<'de> for StringTableIndex {
+    fn deserialize<D: Deserializer<'de>>(deserializer: D) -> Result<Self, D::Error> {
+        let value = String::deserialize(deserializer)?;
```
mstange, line 26: Hmm, why `String::deserialize`? Is it in quotes? It shouldn't be.

samply/src/shared/symbol_precog.rs
```diff
@@ -0,0 +1,386 @@
...
+    fn lookup_sync(&self, address: wholesym::LookupAddress) -> Option<wholesym::SyncAddressInfo> {
+        match address {
+            wholesym::LookupAddress::Relative(rva) => {
+                for (known_rva, info) in &self.known_addresses {
+                    if *known_rva == rva {
+                        //eprintln!("lookup_sync: 0x{:x} -> {}", rva, info.symbol.0);
+                        return Some(wholesym::SyncAddressInfo {
+                            symbol: wholesym::SymbolInfo {
+                                address: rva,
+                                size: None,
```
mstange, line 259: I'd prefer it if you stored the correct symbol address. Sometimes you have many different addresses in the same function, and we want the front-end to create a shared symbol for those different addresses. Here you're giving them all different symbol addresses, so the front-end will think they're different symbols with the same name.

wholesym/src/config.rs
```diff
@@ -20,6 +33,7 @@ pub struct SymbolManagerConfig {
     pub(crate) use_spotlight: bool,
     pub(crate) debuginfod_cache_dir_if_not_installed: Option<PathBuf>,
     pub(crate) debuginfod_servers: Vec<(String, PathBuf)>,
+    pub(crate) precog_data: Option<PrecogDataContainer>,
```
mstange, line 36: As I said in https://github.com/Unity-Technologies/usamply/pull/1#discussion_r1601910497, I would prefer storing it on the helper rather than in the config, and without an extra wrapper struct.

wholesym/src/config.rs
```diff
@@ -1,12 +1,25 @@
...
+    // Explicit implementation needed due to precog_data
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.write_str("...")
+    }
+}
+
 /// The configuration of a [`SymbolManager`](crate::SymbolManager).
 ///
 /// Allows specifying various sources of symbol files.
-#[derive(Debug, Clone, Default)]
```
mstange, line 9: Please keep the config Clone.

mstange (changes requested):

fxprof-processed-profile/src/global_lib_table.rs
vvuk: Yeah, good idea. FWIW, `GlobalLibIndex` is a confusing name, maybe just `UsedLibraryHandle`?

mstange: `UsedLibraryHandle` sounds good

wholesym/src/config.rs
vvuk: Oh, I misunderstood that comment! I missed the part about keeping it in the helper. Will do.

wholesym/src/config.rs
vvuk: Ah yes, this can come back. When I was using a trait object in here, I had all sorts of problems but since that's gone this should be fine. It's moving to the helper anyway.

wholesym/src/config.rs
vvuk: (The wrapper struct was to allow for Debug FWIW -- `SymbolMapTrait` does not imply Debug, so I would have had to define another trait and make sure all symbol map objects where `Debug`, or implement debug manually for `SymbolManagerConfig`)

vvuk:
Ok, implemented proper address/symbol info and properly implemented frames. Should have addressed all the review feedback here. I didn't make the `GlobalLibIndex` -> `UsedLibraryHandle` change, that can be a future thing

mstange (approved):

PR #217, opened 2024-05-20
Implement 'samply setup' to codesign binary

vvuk:


samply/src/mac/process_launcher.rs
```diff
@@ -301,7 +305,7 @@ impl ExistingProcessRunner {
         };
 
         // always root pid first
-        queue_pid(pid, true);
+        queue_pid(pid, false);
```
mstange, line 308: Seems unrelated, is it intentional?

mstange (approved):
Cool! Oh and please fix the typo in the commit message too. :)

samply/src/mac/process_launcher.rs
vvuk: Unrelated yeah, but also intentional; happened to catch it while looking at this and accidentally included it in here. It should be a failure if we can't attach to the root pid, because the profile will be empty.

PR #218, opened 2024-05-22
Don't add markers for unknown ETW events

vvuk:
These are useful for debugging and might be a nice command line option in the future, but for now they just add random `Image/UnLoad` and `Process/Terminated` events which aren't useful

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 target
 profile.json
+profile.syms.json
 .DS_Store
 .direnv/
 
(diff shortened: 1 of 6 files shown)


mstange:
Would you be happy if we just ignored UnLoad and Terminated? I think @jrmuizel likes to experiment with different xperf start settings and then it's convenient to just have the new markers show up immediately.

vvuk:
Yeah, I was actually thinking of just adding a command line flag to make markers for unknown things, because it is pretty useful. Would that work? Specifically:

- Ignore Image/UnLoad and Process/Terminate
- add a `--unknown-event-markers` command line flag to enable generating markers for unknown events

? I'd like an option instead of doing it unilaterally so that if I'm importing an etl file created elsewhere, it may contain lots of events that I don't care about at all, but may end up as a large volume of markers.

vvuk:
Ok, reworked this to add the command line option; let me know if that works.

mstange (approved):

PR #234, opened 2024-05-29
feat: add address to server and import options

Rjected:
Adds the `--address` flag, to set the listening address for the server spun by `record` or `import`. Fixes https://github.com/mstange/samply/issues/86


mstange (approved):
Looks good, thank you!

PR #238, opened 2024-06-03
Fix typo for --windows-symbol-server

ishitatsuyuki:
This almost certainly looks like an unintended typo, but I don't have a Windows binary at hand to test with this.


mstange:
Oops! This was clearly written in a rush and not tested. Thanks!

PR #261, opened 2024-06-18
Add --time-range option for import for Windows

vvuk:
Add a `--time-range` option for `import` on Windows to allow specifying a time range to include in the profile.json from the ETL file. Processing still tracks process/thread/image lifecycle, but samples and events are only added if they're within this time range. This is very helpful in exploring a very large trace to avoid running into profile.json file size issues. (Could be implemented on other platforms as well, though typically the profiles there are much smaller.)

### DIFF
--- a/samply/src/main.rs
+++ b/samply/src/main.rs
@@ -519,13 +519,6 @@ impl ImportArgs {
             let name = self.file.file_name().unwrap_or(self.file.as_os_str());
             name.to_string_lossy().into()
         };
-        cfg_if::cfg_if! {
-            if #[cfg(target_os = "windows")] {
-                let time_range = self.time_range;
-            } else {
-                let time_range = None;
-            }
-        }
         ProfileCreationProps {
             profile_name,
             main_thread_only: self.profile_creation_args.main_thread_only,
@@ -540,7 +533,10 @@ impl ImportArgs {
             unknown_event_markers: self.profile_creation_args.unknown_event_markers,
             #[cfg(not(target_os = "windows"))]
             unknown_event_markers: false,
-            time_range,
+            #[cfg(target_os = "windows")]
+            time_range: self.time_range,
+            #[cfg(not(target_os = "windows"))]
+            time_range: None,
         }
     }
 
@@ -580,22 +576,20 @@ impl RecordArgs {
             std::process::exit(1);
         }
         let interval = Duration::from_secs_f64(1.0 / self.rate);
-        cfg_if::cfg_if! {
-            if #[cfg(target_os = "windows")] {
-                let (vm_hack, keep_etl) = (self.vm_hack, self.keep_etl);
-            } else {
-                let (vm_hack, keep_etl) = (false, false);
-            }
-        }
-
         RecordingProps {
             output_file: self.output.clone(),
             time_limit,
             interval,
-            vm_hack,
             gfx: self.gfx,
             browsers: self.browsers,
-            keep_etl,
+            #[cfg(target_os = "windows")]
+            vm_hack: self.vm_hack,
+            #[cfg(not(target_os = "windows"))]
+            vm_hack: None,
+            #[cfg(target_os = "windows")]
+            keep_etl: self.keep_etl,
+            #[cfg(not(target_os = "windows"))]
+            keep_etl: None,
         }
     }
 
--- a/samply/src/main.rs
+++ b/samply/src/main.rs
@@ -146,6 +146,14 @@ struct ImportArgs {
     /// Enable CoreCLR event conversion.
     #[clap(long, require_equals = true, value_name = "FLAG", value_enum, value_delimiter = ',', num_args = 0.., default_values_t = vec![CoreClrArgs::Enabled])]
     coreclr: Vec<CoreClrArgs>,
+
+    /// Start time to capture samples at, in seconds
+    #[arg(long)]
+    tstart: Option<u32>,
+
+    /// End time to capture samples at, in seconds
+    #[arg(long)]
+    tstop: Option<u32>,
 }
 
 #[allow(unused)]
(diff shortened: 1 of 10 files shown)


vvuk:
Hm how is clippy blowing up on linux for some of this stuff _now_

mstange:
New Rust version came out and added new warnings. I'll fix it and you can rebase.

samply/src/main.rs
```diff
@@ -482,6 +519,13 @@ impl ImportArgs {
             let name = self.file.file_name().unwrap_or(self.file.as_os_str());
             name.to_string_lossy().into()
         };
+        cfg_if::cfg_if! {
+            if #[cfg(target_os = "windows")] {
+                let time_range = self.time_range;
```
mstange, line 524: These braces are lies!

Can you make it a `let mut time_range = None;` at the function scope and then mutate it inside the braces? I find it too confusing otherwise.

samply/src/main.rs
```diff
@@ -537,9 +582,9 @@ impl RecordArgs {
         let interval = Duration::from_secs_f64(1.0 / self.rate);
         cfg_if::cfg_if! {
             if #[cfg(target_os = "windows")] {
-                let vm_hack = self.vm_hack;
+                let (vm_hack, keep_etl) = (self.vm_hack, self.keep_etl);
```
mstange, line 585: Same here, I guess - I didn't realize we were already using this pattern.

samply/src/windows/profile_context.rs
```diff
@@ -315,6 +318,9 @@ impl ProfileContext {
             None
         };
         let main_thread_only = profile_creation_props.main_thread_only;
+        let time_range_ns = profile_creation_props
+            .time_range
+            .map(|(start, end)| (start.as_nanos() as u64, end.as_nanos() as u64));
```
mstange, line 323: Could you convert the range ends to profile Timestamps here? Then you don't need to make the `nanos` member public.

mstange (changes requested):

samply/src/main.rs
vvuk: Yes :) I hate this pattern, would be nice to do something.. nicer.

samply/src/windows/profile_context.rs
vvuk: Oh, yes, good call. Will do. I'll update rust as well, no need to do it

samply/src/main.rs
mstange: Oh and then Rust is going to complain that the mut is unused on non-Windows... you just can't win

philipc: I think this works:
```rust
        let time_range = {
            cfg_if::cfg_if! {
                if #[cfg(target_os = "windows")] {
                    self.time_range
                } else {
                    None
                }
            }
        };
```

vvuk: Nope; that's what I tried first. `Attributes on expressions are experimental.`

vvuk: I'll just write out the expressions for now in the struct initialization. I think in the future we can just add `WindowsImportArgs` or similar that can get flattened in or similar.

samply/src/windows/profile_context.rs
```diff
@@ -535,6 +551,13 @@ impl ProfileContext {
             reference_raw: timestamp_raw,
             raw_to_ns_factor: 1000 * 1000 * 1000 / perf_freq,
         };
+
+        self.time_range = self.time_range_us.map(|(start, end)| {
+            (
+                self.timestamp_converter.convert_us(start),
+                self.timestamp_converter.convert_us(end),
```
mstange, line 558: Doesn't this treat the range as being relative to the QPC origin, rather than relative to profiling start?

mstange: I would have thought you could just use `Timestamp::from_nanos_since_reference` in the constructor. Or do you want to be able to specify ranges relative to the QPC origin?

vvuk: Oops, it should not be `convert_us`, since that's a duration; it should be from_.._since reference. But I can't do it in the ctor -- the timestamp_converter isn't fully set up at that point. If I want to keep these as Timestamps, I need them to have the same origin as the rest of the times.

It also is kind of wonky and unfortunate that ProfileContext's methods are raw timestamps instead of `Timestamp`, since they always come from the event timestamp. Any objections to me fixing?

vvuk: Oh wait, right. I see what you mean. I had timestamp origin confused in my head.

mstange: > It also is kind of wonky and unfortunate that ProfileContext's methods are raw timestamps instead of Timestamp, since they always come from the event timestamp. Any objections to me fixing?

No objections, unless the fix comes with a RefCell

samply/src/main.rs
philipc: It doesn't matter now, but it does work if you include the extra braces, like I showed, since that converts it into a statement. Omitting those will give the error you saw.

samply/src/windows/etw_gecko.rs
```diff
@@ -116,6 +116,10 @@ pub fn profile_pid_from_etl_file(context: &mut ProfileContext, etl_file: &Path)
                 // nothing, but we don't want a marker for it
             }
             "MSNT_SystemTrace/StackWalk/Stack" => {
+                if !context.is_in_time_range(timestamp_raw) {
```
mstange, line 119: I wonder if this should use referenced_timestamp_raw instead - that's what we use for the sample timestamp.

mstange (approved):
r+

mstange:
Merging so that we don't have to keep rebasing it. We can make any tweaks as follow-ups.

PR #291, opened 2024-07-13
Improve handling of unmerged ETL files.

mstange:
We still merge during recording. But this is one step on the way to a world where we no longer need to do it. We could probably stop merging right now for the case where we only have a kernel session and no user session.

This commit tweaks how we buffer the ImageID information before we process the MSNT_SystemTrace/Image information.
And it adds handling to get the CodeId + DebugId from the binary, if we're importing an unmerged ETL file.

I've removed a fair amount of code that we had kept around for adding kernel driver handling. It doesn't seem to be needed - I get MSNT_SystemTrace/Image/DCStart events for all the same kernel modules that I also see in the merged ETL.

@vvuk Do you think we still need any of the code in winutils.rs?


mstange:
> @vvuk Do you think we still need any of the code in winutils.rs?

Oh, we still call `get_dos_device_mappings()`, so that part needs to stay. But I think `iter_kernel_drivers()` can go. And `enable_debug_privilege()` should go anyway.

vvuk:
This looks good. And agreed that `iter_kernel_drivers()` + `enable_debug_privilege()` can go away. For debug privilege, I can maybe see a future world where samply may want to inspect a target process directly, but we can bring enable privs when needed.

mstange:
Thanks for confirming!

PR #310, opened 2024-07-24
Update dependencies.

mstange:


mstange:
I've filed https://github.com/gimli-rs/addr2line/issues/311 on the test failure and will skip the addr2line/gimli update for now.

PR #358, opened 2024-09-16
Use Rust 1.80 until aarch64 test failures are fixed with 1.81

vvuk:
The linux cross compilation tests have failures with 1.81, and it's nothing obvious at quick glance from the (massive) log output. This explicitly uses Rust 1.80 for the cross tests until someone can take a look at the 1.81 failures.


vvuk:
Note: `Rust / build & test (aarch64-apple-darwin) (pull_request)` just failed; I'm re-running but I think it failed the same way as the cross tests did. It's still using latest rust stable (1.81) instead of 1.80.

PR #363, opened 2024-09-27
Detect python functions by the "py::" prefix.

mstange:
Python has support for perf JIT maps. It prefixes the python function names with "py::": https://github.com/python/cpython/blob/4b89c5ebfc7d5d4f008eee0ae6da765dfc28e3a9/Python/perf_trampoline.c#L218

To use this, on Linux, build python 3.12 with frame pointers enabled, and run your python script with the `PYTHONPERFSUPPORT=1` environment variable set:

```
CFLAGS="-fno-omit-frame-pointer -mno-omit-leaf-frame-pointer" pyenv install 3.12
pyenv shell 3.12
samply record PYTHONPERFSUPPORT=1 python stuff.py
```


ngoldbaum:
This is extremely useful, thanks for adding this!!

PR #403, opened 2024-10-31
Use `mach_vm_read_overwrite` to read image infos for dyld_images

Maaarcocr:
As I've explained in https://github.com/mstange/samply/issues/389#issuecomment-2450297577 it seems like the mapped memory that samply relies on to read image infos for dyld_images is no longer kept up to date with the newest version of macOS. 

This results in samply not being able to reconcile from where an address is coming from (as it was shown in the issue I already linked) and thus then we get no symbolication whatsover. 

I have a simple solution for this (which I understand may not be desirable from a perf POV): for image infos just read them directly from the other process memory using `mach_vm_read_overwrite`. This may be slower, but it at least get samply working again when people use dlopen in the application code. Also, even if it is slower, hopefully people are not constantly dlopen-ing new libraries so this code should only be called few times at startup and that should be it hopefully?

### DIFF
--- a/samply/src/mac/proc_maps.rs
+++ b/samply/src/mac/proc_maps.rs
@@ -1,5 +1,6 @@
 use std::cmp::Ordering;
 use std::collections::HashMap;
+use std::mem::MaybeUninit;
 use std::ops::{Deref, Range};
 use std::{mem, ptr};
 
@@ -21,7 +22,7 @@ use mach::thread_act::{thread_get_state, thread_resume, thread_suspend};
 use mach::thread_status::thread_state_flavor_t;
 use mach::thread_status::thread_state_t;
 use mach::traps::mach_task_self;
-use mach::vm::{mach_vm_deallocate, mach_vm_read, mach_vm_remap};
+use mach::vm::{mach_vm_deallocate, mach_vm_read, mach_vm_read_overwrite, mach_vm_remap};
 use mach::vm_inherit::VM_INHERIT_SHARE;
 use mach::vm_page_size::{mach_vm_trunc_page, vm_page_size};
 use mach::vm_prot::{vm_prot_t, VM_PROT_NONE, VM_PROT_READ};
@@ -158,6 +159,7 @@ impl DyldInfoManager {
 
             let new_image_info = enumerate_dyld_images(
                 &mut self.memory,
+                self.task,
                 info_array_addr,
                 info_array_count,
                 dyld_image_load_addr,
@@ -211,6 +213,7 @@ fn with_suspended_task<T>(
 
 fn enumerate_dyld_images(
     memory: &mut ForeignMemory,
+    task: mach_port_t,
     info_array_addr: u64,
     info_array_count: u32,
     dyld_image_load_addr: u64,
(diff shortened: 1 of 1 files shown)


samply/src/mac/proc_maps.rs
```diff
@@ -227,8 +230,17 @@ fn enumerate_dyld_images(
         let (base_avma, image_file_path) = {
             let info_array_elem_addr =
                 info_array_addr + image_index as u64 * mem::size_of::<dyld_image_info>() as u64;
-            let image_info: &dyld_image_info =
-                unsafe { memory.get_type_ref_at_address(info_array_elem_addr) }?;
+            let mut image_info: MaybeUninit<dyld_image_info> = MaybeUninit::uninit();
+            let mut size = mem::size_of::<dyld_image_info>() as u64;
+            unsafe {
+                mach_vm_read_overwrite(task, info_array_elem_addr, size, &mut image_info as *mut MaybeUninit<dyld_image_info> as u64, &mut size).into_result()?;
```
ianks, line 236: nit:

```suggestion
                mach_vm_read_overwrite(task, info_array_elem_addr, size, &mut image_info as *mut MaybeUninit<dyld_image_info> as mach_vm_address_t, &mut size).into_result()?;
```

samply/src/mac/proc_maps.rs
```diff
@@ -227,8 +230,17 @@ fn enumerate_dyld_images(
         let (base_avma, image_file_path) = {
             let info_array_elem_addr =
                 info_array_addr + image_index as u64 * mem::size_of::<dyld_image_info>() as u64;
-            let image_info: &dyld_image_info =
-                unsafe { memory.get_type_ref_at_address(info_array_elem_addr) }?;
+            let mut image_info: MaybeUninit<dyld_image_info> = MaybeUninit::uninit();
+            let mut size = mem::size_of::<dyld_image_info>() as u64;
```
ianks, line 234: nit:

```suggestion
            let mut size = mem::size_of::<dyld_image_info>() as mach_vm_size_t;
```

samply/src/mac/proc_maps.rs
```diff
@@ -227,8 +230,17 @@ fn enumerate_dyld_images(
...
+            unsafe {
+                mach_vm_read_overwrite(task, info_array_elem_addr, size, &mut image_info as *mut MaybeUninit<dyld_image_info> as u64, &mut size).into_result()?;
+            }
+            
+            if size != mem::size_of::<dyld_image_info>() as u64 {
```
ianks, line 239: We should also check that the result of `mach_vm_read_overwrite ` was `KERN_SUCCESS`, I think

casperisfine:
I just tried this, and while it now include symbols, they don't quite look correct
Capture d’écran 2024-11-04 à 10 42 33

samply/src/mac/proc_maps.rs
Maaarcocr: into_result does it

Maaarcocr:
> I just tried this, and while it now include symbols, they don't quite look correct Capture d’écran 2024-11-04 à 10 42 33

could you try again? I've pushed some new changes that make us of `mach_vm_read_overwrite` more and it seems to work for my native gem!

casperisfine:
It works now ! ❤️ 

Capture d’écran 2024-11-04 à 12 39 44

mstange:
Thank you @Maaarcocr for the patch, @ianks for the review and @casperisfine for testing!

PR #416, opened 2024-11-16
Skip remaining iterations on Windows too

Rawk:
On Linux and Mac, remaining iterations are skipped if the command returns non-success exit status. Do the same on Windows.


mstange:
Thanks!

PR #425, opened 2024-11-18
Profile child processes on macOS

### LINKED ISSUE #419: option to attach to a process tree
would be nice if `samply` could take multiple `-p` options, or ideally be able to follow and attach to all child processes of the pid passed in

tmm1:
closes #419


mstange:
Thanks!

PR #490, opened 2025-02-06
Add ignore exit code flag

kaleidawave:
Add `ignore_exit_code` flag so that iteration count can continue even if the process returned a non-zero exit code (which is common on linters / verifiers etc)

### DIFF
--- a/samply/src/windows/profiler.rs
+++ b/samply/src/windows/profiler.rs
@@ -100,10 +100,6 @@ pub fn start_recording(
 
                 pids.push(child.id());
 
-                if process_launch_props.ignore_exit_code {
-                    continue;
-                }
-
                 // Wait for the child to exit.
                 //
                 // TODO: Do the child waiting and the xperf control on different threads,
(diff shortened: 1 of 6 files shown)


samply/src/windows/profiler.rs
```diff
@@ -100,6 +100,10 @@ pub fn start_recording(
 
                 pids.push(child.id());
 
+                if process_launch_props.ignore_exit_code {
```
mstange, line 103: I don't think this is right - you need to wait for the child process to finish before you launch the next iteration. Can you combine this check with the existing `!exit_status.success()` check?

mstange (changes requested):

samply/src/windows/profiler.rs
kaleidawave: You are correct, this condition should go after the `wait` call. Have updated 👍

mstange (approved):
Looks good to me!

kaleidawave:
Have put in the fix for `mac`, so should be okay now

mstange:
Merged, thanks!

PR #542, opened 2025-04-11
Enable --unstable-presymbolicate for 'samply import'

manh9203:
`samply import` currently supports an `--unstable-presymbolicate` option but does not output the `.syms.json` file when this flag is enabled. This PR adds the missing code to generate the symbol JSON file, which is useful for running in CI.


mstange:
Thanks, can you add the equivalent code to the ETL import code path as well?
https://github.com/mstange/samply/blob/a1a651ef087d9652cc8b9e1fcd32d91722f81f92/samply/src/windows/import.rs#L38

manh9203:
> Thanks, can you add the equivalent code to the ETL import code path as well?
> 
> https://github.com/mstange/samply/blob/a1a651ef087d9652cc8b9e1fcd32d91722f81f92/samply/src/windows/import.rs#L38

I just added it.

mstange:
Thank you! I won't have time for it tonight but hopefully I can find some time over the weekend to make a new release.

PR #567, opened 2025-05-13
suggest -1 for perf_event_paranoid

sylvestre:
To workaround:
Failed to start profiling: mmap failed
https://github.com/mstange/samply/issues/25


mstange:
Thank you!

PR #606, opened 2025-06-26
Enhance PerfBuilder::open error messages with last_os_error

DaniPopes:
Add the specific OS error message to 'mmap failed' in PerfBuilder::open.

Also use it above when checking the result of [perf_event_open](https://www.man7.org/linux/man-pages/man2/perf_event_open.2.html):
> RETURN VALUE
>
> On success, perf_event_open() returns the new file descriptor.  On
> error, -1 is returned and [errno](https://www.man7.org/linux/man-pages/man3/errno.3.html) is set to indicate the error.


mstange:
Thank you!

PR #613, opened 2025-07-01
Add Flow related formats

jrmuizel:


mstange:
Actually it might be best to not treat it as a string, and instead call a new `flow_field_value` method for these fields

jrmuizel:
This version is a lot more involved but seems necessary. It was mostly written by Cursor.

fxprof-processed-profile/src/marker_table.rs
```diff
@@ -199,6 +218,12 @@ impl Serialize for SerializableMarkerDataElement<'_> {
                     (value, number_fields) = number_fields.split_first().unwrap();
                     map.serialize_entry(&field.key, value)?;
                 }
+                MarkerFieldFormatKind::Flow => {
+                    let value;
+                    (value, flow_fields) = flow_fields.split_first().unwrap();
+                    // Flow IDs are serialized as hex strings without 0x prefix
+                    map.serialize_entry(&field.key, &format!("{:x}", value))?;
```
mstange, line 225: This needs to actually use the string table and output the string index.

fxprof-processed-profile/tests/integration_tests/main.rs
```diff
@@ -1533,3 +1541,97 @@ fn profile_counters_with_sorted_processes() {
...
+        MarkerTiming::Instant(Timestamp::from_millis_since_reference(10.0)),
+        flow_marker,
+    );
+
+    // Serialize to JSON and check the output
+    let json_string = serde_json::to_string_pretty(&profile).unwrap();
+
+    // The marker should contain our flow IDs serialized as hex strings
+    assert!(json_string.contains("\"flowId\": \"ab54a98ceb1f0ad2\""));
+    assert!(json_string.contains("\"termFlowId\": \"891087b8e3b70cb1\""));
```
mstange, line 1632: Just use `assert_json_eq!` here like in the other tests. Or you can ask Cursor to convert it to use the insta crate, if insta can do JSON snapshots. (I haven't checked if it can.)
And then make the test check that it's actually using string indexes and not strings

mstange (changes requested):
looks much better, two comments left, and please make sure clippy passes with current stable rust

fxprof-processed-profile/src/markers.rs
```diff
@@ -90,6 +90,21 @@ pub trait Marker {
...
+    ///
+    /// Flow identifiers are u64 values that are unique across processes.
+    ///
+    /// You can panic for any unexpected field indexes, for example
+    /// using `unreachable!()`. You can even panic unconditionally if this
+    /// marker type doesn't have any flow fields.
+    ///
+    /// If you do see unexpected calls to this method, make sure you're not registering
+    /// multiple different schemas with the same [`RuntimeSchemaMarkerSchema::type_name`].
+    fn flow_field_value(&self, field_index: u32) -> u64;
```
mstange, line 107: Should we use a `FlowId(u64)` newtype here? Not sure if that would help any. But it might provide a surface for more doc comments

fxprof-processed-profile/src/marker_table.rs
mstange: Ah, it probably needs to happen in `add_marker`, where we call `flow_field_value`.

jrmuizel: So add_marker would need a mut reference to the Profile then?

jrmuizel:
Here's a version that does the string interning. I haven't looked at it closely

jrmuizel:
Do you feel good about this approach? I'm not loving it.

mstange:
It's not super pretty but that's the only idea I have, given the requirements for the output. It seems good enough for now.

jrmuizel:
Ok, I'll polish it up

mstange (approved):
Thanks!

PR #685, opened 2025-10-08
Split out samply-debugid crate

### LINKED ISSUE #684: Split out debugid_utils into crate
Hi, I've put together a profiler for bare-metal embedded systems that I'm planning to contribute to [probe-rs](https://github.com/probe-rs/probe-rs). Currently I can record my profiles and write them out using `fxprof-processed-profile` with my `probe-rs` branch and then open them with `samply load` . To do this though I've copied and pasted the contents of [debug_util.rs](https://github.com/mstange/samply/blob/80859bb325f3129343e1052ffc6d681a0a248395/samply-symbols/src/debugid_util.rs), since `samply` needs the `breakpadId` to match.

Would a PR to separate the contents of `debug_util.rs` into a `samply-debugid` crate be accepted? Then I could re-use this code without pulling in the rest of the dependencies of `samply-symbols` which I don't need.

KingCol13:
Fixes #684 

Looking through `samply-symbols/src/shared.rs` I was wondering if there's anything else worth breaking out for use when writing profiles. For example I used the ELF part of `relative_base_address` (although this is quite small so I'm not sure this function in particular is worth it).


mstange:
Thank you! I agree, relative_base_address is the one other thing that came to mind. Maybe we should also add a samply-object crate, and then move debug_id_for_object and relative_address_base into it? Then samply-debugid doesn't need a dependency on object. We could even make samply-object define an ObjectExt trait if we wanted.

mstange:
I'll merge this as-is, but let me know your thoughts.

KingCol13:
> Maybe we should also add a samply-object crate, and then move debug_id_for_object and relative_address_base into it? Then samply-debugid doesn't need a dependency on object. We could even make samply-object define an ObjectExt trait if we wanted.

I like this idea, I'll open a PR.

PR #689, opened 2025-10-19
Split out samply-object crate

KingCol13:
Builds on #685 from discussion [here](https://github.com/mstange/samply/pull/685#issuecomment-3418635426)

- Remove `debug_util.rs` (somehow I missed this in #685)
- Moves `code_id_from_object` and `debug_id_from_object` into ObjectExt trait in `samply-object`
- Moves `relative_address_base` into `ObjectExt::samply_relative_address_base`
- Removes some related re-exports in `samply-symbols` (breaking change)

I'm ambivalent as to whether `ObjectExt` exists or we just use free-standing functions but this does match the style of the codebase and maybe does make it more ergonomic.

The `samply_relative_address_base` name seems error prone when `Object::relative_address_base` exists. I don't think I understand relative base addresses and different executable formats well enough to suggest a better name/solution. I certainly wonder if it would be useful for `Object`'s method to use this implementation at least for ELFs.

### DIFF
--- a/samply-symbols/src/lib.rs
+++ b/samply-symbols/src/lib.rs
@@ -211,7 +211,7 @@ use jitdump::JitDumpIndex;
 use linux_perf_data::jitdump::JitDumpReader;
 use object::read::FileKind;
 pub use pdb_addr2line::pdb;
-pub use samply_debugid::{CodeId, ElfBuildId, PeCodeId};
+pub use samply_debugid::{CodeId, DebugIdExt, ElfBuildId, PeCodeId};
 pub use samply_object::{debug_id_for_object, relative_address_base};
 use shared::FileContentsCursor;
 pub use {debugid, object};


mstange:
Ah, I forgot about the object function of the same name. In that case, free standing functions seem preferable over an extension trait.

KingCol13:
> Ah, I forgot about the object function of the same name. In that case, free standing functions seem preferable over an extension trait.

Yea I was beginning to lean back that way, I'll swap them back.

KingCol13:
Ok, they're back to being free standing functions. I added the re-exports back so there shouldn't be any breaking changes to `samply-symbols` any more.

mstange:
Perfect, thank you!

philipc:
>  I certainly wonder if it would be useful for Object's method to use this implementation at least for ELFs.

`Object::relative_address_base` is intended to be something that you can add to things like `ObjectSymbol::address` to get a SVMA, without needing to care what object file format you are using. Maybe `ObjectSymbol::address` should have had this base automatically added in, but I think it's too late for that now. My understanding is that using the first load segment address as a relative address base for ELF and Mach-O is a Breakpad convention, but I'm not sure; I'm interested in learning more. I think most of the ELF world works with SVMAs, not relative addresses.

philipc:
Oh, I misremembered. `ObjectSymbol::address` does already add in the base. So maybe `Object::relative_address_base` is only the way it is because [`virtual_memory_bias` in `findshlibs` is inconsistent](https://github.com/gimli-rs/findshlibs/issues/76).

PR #690, opened 2025-10-20
Extract the markName and measureName fields from performance.measure …

dpalmeiro:
The patch in https://source.chromium.org/chromium/chromium/src/+/453a29751849863627f6cabfe2645dd48c8d6c33 changed how Chrome emits the measure and mark events for ETW.  The name of the event is now stored in the `measuredName` and `markName` fields now.  For measure events, the name is only stored in the BEGIN phase.


mstange:
Do you have an example profile that was imported with this patch?

dpalmeiro:
Here is a sample profile using Canary:  https://share.firefox.dev/42NV47E

dpalmeiro:
image
Attached is a screenshot of the ETL file for some measure events when viewed from WPA.  (The `Id` and `Timestamp` fields are swapped between Begin & End events so it looks a bit confusing here.)

PR #710, opened 2025-11-11
Add samply-markers crate

nordzilla:
## Overview

This is a minimal initial implementation of #349 based off of @blp's commit https://github.com/feldera/feldera/commit/1b448b757e92c245a02be31932278871cf940a65.

I've pulled the commit directly into my pull request, so as to preserve the original attributions, and built the crate around the Unix mmap capabilities. Windows is stubbed out, but not yet implemented.

## Examples

The [README](https://github.com/nordzilla/samply/blob/samply-markers/samply-markers/README.md) and generated rust docs are full of examples to look at.

## Crate Name Reservation

I've also reserved the crate name in good faith that I'd like to release this eventually.
* https://crates.io/crates/samply-markers

Though, if for some reason this falls through, I'd be happy to relinquish ownership of the crate. I'm not a fan of name squatting without substance or genuine intent.

## Current Status

That being said, I'm not ready to release the crate as `0.1.0` yet. This is just the first pull request.

This implementation does the bare minimum that is (mostly) compatible with the current `samply` implementation (I did have to make one small change that I would prefer to revert in time).

**What's implemented:**
* [x] The core public API that I'm fairly committed to, though welcome to feedback.
* [x] Full Unix support (Linux and macOS). 
* [x] Documentation with links to Firefox Profiler examples.
* [x] Runnable examples demonstrating synchronous, asynchronous profiling.

## Future Work

The following are all considerations that I _would like_ to handle before a `0.1.0` crate release, but aren't necessarily critical functionality. 

### Windows Support

Windows is currently stubbed out with `unimplemented!()`. The implementation will need platform-specific equivalents for the Unix mmap-based approach.

> [!NOTE]
> Of all the topics, I'm most okay with releasing `0.1.0` without this. 
> I don't have a Windows dev machine to implement or test this on. 

### Runtime Detection

I would like to have runtime detection for emitting markers if possible. Perhaps samply itself could set `SAMPLY_MARKERS=1` for the process to run, or if the user already has `SAMPLY_MARKERS=0` in the current session, then Samply could preserve that.

That would be a lot more ergonomic to use, since markers would either emit if samply is recording, or not if samply is not, and likely not cost much performance compared to the currently implemented compile-time disablement.

### Extended Marker Data

I would like to allow `samply-markers` to output more optional data that can take advantage of runtime marker schemas, but this will require more changes to Samply itself.

Ideally, I'd like to redo the format and parsing of the marker files (which would be a breaking change).

I could do this in a backward-compatible, or perhaps I could make my own format that outputs to `samply-marker-{PID}-{TID}.txt`.

I'm not sure how important it is that the current format remains intact.

## Known Issues

The following are all considerations that will need to be handled before a `0.1.0` crate release. But I don't think they should block merging this PR, creating these as sub-issues, and getting this crate hooked up to CI for testing.

### Buffer Flushing

Currently, the marker-file buffers use thread-local storage (TLS) and ensure flushing on drop. This approach doesn't work for long-lived threads, such as those in Rayon's global thread pool. As a result, buffered marker data may never be written out for long-lived threads.

I really don't want to have to mutex-guard my writes to the buffers, but at this point, I can't think of a safe solution that isn't a centralized flush mechanism on the main thread. 

### Buffer Size

The current implementation has a 64 KiB buffer per thread to write markers, because it performed well in the single not-very-principled hyperfine benchmark that I ran. But 64 KiB feels like a lot. I'll definitely need to do more research and benchmarking into what is an ideal buffer size to have for each thread, both in terms of size and speed. 

### Ctrl+C Handling

This issue is related to buffer flushing. Whatever solution is developed for the flushing problem will likely be used here as well, but explicit handling of `Ctrl+C` will still be needed to ensure that buffers are properly flushed before program termination.

### Attaching to a Process with Markers

Markers are emitted from the process that is being recorded, and this can make things look akward in the profiler UI when samply is attached to an already-running process. It means that the marker start time for some intervals could be earlier than the profiling start time. 

I think the cleanest solution here would be to filter markers as they are added to the JSON, to ensure that any interval markers' start times are equal to or after the profiling start time. 

This would require Samply to know the start time of the PID it just attached to. There may be system calls to find this information (I haven't looked too deeply yet), or if we revise the marker file format, we could ensure that every marker file includes a global first time-since-epoch-style timestamp that is guaranteed to be earlier than the first marker. This would allow Samply to reason about the marker timestamp data more easily. 

---

@mstange, perhaps we could collaborate on this in a new issue once this stack is reviewed and (hopefully) landed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2462,6 +2462,7 @@ name = "samply-markers"
 version = "0.1.0"
 dependencies = [
  "futures",
+ "itoa",
  "libc",
  "mach2 0.4.3",
  "nix",
@@ -2469,7 +2470,6 @@ dependencies = [
  "regex",
  "reqwest",
  "serial_test",
- "smallstr",
  "tempfile",
  "tokio",
 ]
@@ -2699,15 +2699,6 @@ name = "slab"
 version = "0.4.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7a2ae44ef20feb57a68b23d846850f861394c2e02dc425a50098ae8c90267589"
-
-[[package]]
-name = "smallstr"
-version = "0.3.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "862077b1e764f04c251fe82a2ef562fd78d7cadaeb072ca7c2bcaf7217b1ff3b"
-dependencies = [
- "smallvec",
-]
 
 [[package]]
 name = "smallvec"
--- a/samply-markers/Cargo.toml
+++ b/samply-markers/Cargo.toml
@@ -14,14 +14,14 @@ keywords = ["profiling", "markers", "instrumentation", "tracing", "performance"]
 [features]
 default = []
 enabled = [
+    "dep:itoa",
     "dep:nix",
-    "dep:smallstr",
     "dep:tempfile",
 ]
 
 [target.'cfg(unix)'.dependencies]
+itoa = { version = "1.0.15", optional = true }
 nix = { version = "0.30.1", features = ["mman", "process", "time"], optional = true }
-smallstr = { version = "0.3.1", optional = true }
 tempfile = { version = "3.23.0", optional = true }
 
 [target.'cfg(target_os = "macos")'.dependencies]
--- a/samply-markers/src/marker/timestamp.rs
+++ b/samply-markers/src/marker/timestamp.rs
@@ -38,16 +38,14 @@ impl SamplyTimestamp {
         Self(nanos)
     }
 
-    /// Formats a [`SamplyTimestamp`] without implementing [`std::fmt::Display`].
+    /// Returns the inner nanosecond value.
     ///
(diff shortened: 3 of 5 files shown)


blp:
What a gift! Thank you!

(But I haven't looked beyond the commit message.)

nordzilla:
> What a gift! Thank you!
> 
> (But I haven't looked beyond the commit message.)

Please give it a try. I'd appreciate your feedback.

It's slightly different than what you implemented, and I've removed the dependency on `tracing`. 

But I'm hoping the documentation generated by `cargo doc`, as well as in the [README](https://github.com/nordzilla/samply/blob/samply-markers/samply-markers/README.md), is easy to follow, and that the design/intent is intuitive.

DaniPopes:
Hey, just wanted to flag that I wrote a `tracing` adapter for the marker files that people might prefer using since they wouldn't have to write any new "spans": https://github.com/DaniPopes/tracing-samply/ (previous art: [tracing-chrome](https://docs.rs/tracing-chrome), [tracing-tracy](https://docs.rs/tracing-tracy))

PR #729, opened 2025-12-02
Add support for interposing mkostemps on macOS

justinmichaud:
This is the easiest way to make files without requiring a sandbox exception. I only added mkostemps, and not others, simply because that is what WebKit uses and it is easy to test.

This also includes a drive-by build fix by including c_int.


justinmichaud:
macOS is really not happy about fsyncing a temporary file, and I figured out the issue preventing this from working in /tmp in WebKit

mstange:
I'm not familiar with `mkostemps`, but if this allows using samply without having to disable the sandbox then it still seems useful. Or are you saying that writing to /tmp is allowed even with the sandbox enabled, as long as you don't fsync?

justinmichaud:
I tried using mkostemps to get around the sandbox, but found that the files it created could not be fsynced. In the end, it made more sense to extend the sandbox to let every process write to /tmp. This was already the case for most processes, the difference is that I was using MiniBrowser instead of Safari to test. With that in place, everything works perfectly in /tmp!

mstange:
Ah, nice.

PR #730, opened 2025-12-12
profiler: check paranoia level after trying SwCpuClock

abhizer:
In some setups, like running in an EKS cluster, currently there's a false negative where the profiler exits before trying SwCpuClock because it sees that the `perf_event_paranoid` level is set to 2. However, if the binary has been given correct capability (CAP_PERFMON), the paranoia level doesn't really matter.

This patch fixes it to only check the paranoia level and error if we fail to run with both Hardware and Software CPU Clock events.


mstange:
Ugh committed my suggestion badly. Will make a new PR.

mstange:
#751

PR #748, opened 2026-02-03
fix(linux): `--reuse-threads` thread naming and recycling

shekhirin:
Two issues were causing incorrect behavior with `--reuse-threads` on Linux:

1. In `handle_fork`, passing the parent thread's name caused all threads to attempt recycling with the same name at creation time, before they got their actual names via COMM events. Now we pass None - threads get their names and recycle properly when the COMM event arrives.

2. In `rename_non_main_thread`, when recycling was enabled but no recycled thread with the target name existed, the rename was silently skipped. Added else branch to fall back to `rename_without_recycling`.

| Before | After |
|--------|--------|
| image | image |


mstange:
Nice, thanks! Can you rebase and run `cargo fmt`?

PR #767, opened 2026-03-25
fix: respect --duration time limit on Linux

DaniPopes:
The `--duration` flag had no effect on Linux. The `time_limit` parameter was passed into `run_profiler` but never checked (prefixed with `_`). This adds an elapsed time check at each loop iteration, matching the existing macOS behavior in `sampler.rs`.


mstange (commented):
Hmm, is there a way to get into a state where there are buffered samples in the event buffer, and they're within the desired time window, but we discard them because we exit this loop before we consume those samples?

samply/src/linux/profiler.rs
```diff
@@ -595,87 +679,15 @@ fn run_profiler(
...
-                pending_lost_events = 0;
-            }
-        });
+        perf.consume_events(&mut handle_event);
 
         perf.wait();
     }
 
+    // Flush any events still buffered in the sorter's heap that were held back
+    // waiting for the next round that never came.
```
mstange, line 688: Ok I hadn't actually thought of events stuck in the sorter but that's a good point.

But my question was about events in the perf group members, i.e. the things we access via `perf.are_events_pending()` and `perf.iter()`. Couldn't there be pending events in there?

DaniPopes: It should be handled by the last `consume_events` call right? (in this line of code)

mstange: Oops, yes, I missed that call. Ok good, thanks!

PR #769, opened 2026-03-28
Add missing gather_context_switches to many-threads fallback path

DaniPopes:
When `cpu_count * (threads + 1) >= 1000`, `open_process` falls back to per-thread `any_cpu()` perf events instead of per-cpu-per-thread events to reduce fd count. This fallback path was missing `gather_context_switches()`.

Without context switch events, `consume_cpu_delta` returns the full inter-sample interval as on-CPU time, making every thread appear at 100% CPU utilization in the profiler even when idle.

This also extracts the common builder setup into a closure to prevent future divergence. Note that `inherit_to_children` is only set for the per-cpu path since it's incompatible with `any_cpu()` (`perf_event_open` returns an error when both inherit and cpu==-1 are set).


mstange (approved):
Thanks!

PR #784, opened 2026-04-07
Preserve DWARF column numbers in convert_stack_frame

### LINKED ISSUE #783: samply-symbols drops DWARF column numbers in convert_stack_frame
Unless I am missing some fundamental thing seems that DWARF column numbers are present in the line table but get dropped in `samply-symbols` DWARF conversion path. I traced it to the fact that `convert_stack_frame` in `samply-symbols/src/dwarf.rs` copies `line_number` from`addr2line::Frame::location` but does not copy `location.column`, even though `FrameDebugInfo` already has a `column_number` field.

pablogsal:
convert_stack_frame copied line_number from addr2line::Frame::location but dropped location.column, even though FrameDebugInfo already has a column_number field. This caused downstream symbolizers to lose native source columns despite DWARF containing them.

Fix by reusing the resolved location, forwarding location.column into FrameDebugInfo::column_number, and adding a unit test that covers the conversion path directly.

Fixes #783


samply-symbols/src/dwarf.rs
```diff
@@ -49,20 +50,54 @@ pub fn convert_stack_frame<R: gimli::Reader>(
...
         .as_ref()
         .and_then(|l| l.file)
         .map(|file| string_interner.intern_owned(file).into());
 
     FrameDebugInfo {
         function,
         file_path,
-        line_number: frame.location.and_then(|l| l.line),
+        line_number: location.as_ref().and_then(|l| l.line),
+        column_number: location.as_ref().and_then(|l| l.column),
```
mstange: Does the meaning of the column match here? FrameDebugInfo wants 1 for the beginning of the line, I think dwarf might use 0 for that? But I'm not sure. Do you have an example binary with dwarf that has non-zero columns?

mstange (commented):
Sorry, looks like I never submitted the review comment I had written :(

samply-symbols/src/dwarf.rs
pablogsal: Good catch. I consulted the DWARF 5 standard (https://dwarfstd.org/doc/DWARF5.pdf) and seems that gimli / DWARF uses column 0 for the left edge so I will normalize that case when copying the value. Non-zero DWARF columns do exist in practice as well.

mstange: Should it be + 1 rather than .max(1)? Which compilers output DWARF column numbers?

pablogsal: Maybe I am misunderstanding this but I tried this with a real GCC-produced binary locally with thsi example C source:

  ```c
  __attribute__((noinline)) int multi_stmt(int a) {
      int x = a + 1; int y = x + 2; return y * 3;
  }

  int main(void) {
      return multi_stmt(4);
  }
```

  Built with:

  ```
  gcc -g -O0 -fno-inline -no-pie multi_stmt.c -o multi_stmt_gcc
  ```

  This is the raw DWARF line table (  `llvm-dwarfdump --debug-line multi_stmt_gcc`):

```
  Address            Line   Column File   ISA Discriminator OpIndex Flags
  0x0000000000401106      1     49      1   0             0       0  is_stmt
  0x000000000040110d      2      9      1   0             0       0  is_stmt
  0x0000000000401116      2     24      1   0             0       0  is_stmt
  0x000000000040111f      2     44      1   0             0       0  is_stmt
  0x0000000000401128      3      1      1   0             0       0  is_stmt
```

I also wrote a tiny Rust program that reads the binary with `addr2line::Loader` and prints the observed locations. Its output for the same binary was:

```
  0x00401106: /tmp/dwarf-column-demo/cases/multi_stmt.c:1:49
  0x0040110d: /tmp/dwarf-column-demo/cases/multi_stmt.c:2:9
  0x00401116: /tmp/dwarf-column-demo/cases/multi_stmt.c:2:24
  0x0040111f: /tmp/dwarf-column-demo/cases/multi_stmt.c:2:44
  0x00401128: /tmp/dwarf-column-demo/cases/multi_stmt.c:3:1
```

So, based on this example, GCC is definitely producing real non-zero DWARF column numbers, and addr2line is returning those same values unchanged.

 I also tried rebuilding the same file with `-gno-column-info`, and then the observed columns were 0:

```
  0x00401106: /tmp/dwarf-column-demo/cases/multi_stmt.c:1:0
  0x0040110d: /tmp/dwarf-column-demo/cases/multi_stmt.c:2:0
  0x00401128: /tmp/dwarf-column-demo/cases/multi_stmt.c:3:0
```

The dwarf 5 standard says:

> (columns are) An unsigned integer indicating a column
number within a source line. Columns are
numbered beginning at 1. The value 0 is
reserved to indicate that a statement begins at
the “left edge” of the line

So my current understanding is that I do not think `+ 1` is right here, because that would turn a real DWARF column like 9 into 10. It looks more like DWARF columns are already 1-based when they are real columns, and 0 is the special sentinel case so given our `FrameDebugInfo` contract is 1-based, `.max(1)` seems closer to what we want: it preserves real non-zero columns and only normalizes the 0 case.

Apologies if I am missing something here

mstange: Ah perfect, thanks for the example! So then I agree that `.max(1)` is right. Would you mind adding a comment along the lines of `In DWARF, column 0 and 1 both refer to the start of the line, with 0 sometimes indicating the absence of column information`?
Actually then I wonder if we should set the column to None if it's zero... your call.

pablogsal: > Actually then I wonder if we should set the column to None if it's zero... your call.

I think I prefer to keep it as  0 is not “unknown” in the spec; it is a real “left edge of line” marker. Since we can’t distinguish “true left edge” from “compiler omitted precise columns and emitted 0”, converting to None would throw away valid start-of-line locations. Converting to 1 preserves the closest representable 1-based source location IMHO

mstange: Makes sense, thanks! I'm still a bit confused what they mean by left edge and how it's different from start-of-line aka column 1. Anyway, merging.

PR #804, opened 2026-05-04
Enable support for profiles created with Simpleperf's on-the-fly compression (-z)

animalan:
Allow Samply to process Simpleperf profiles created with the `-z` compression option.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1489,6 +1489,7 @@ dependencies = [
  "prost",
  "prost-derive",
  "thiserror 2.0.18",
+ "zstd-safe",
 ]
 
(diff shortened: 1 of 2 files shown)


mstange:
Thanks!

mstange:
Looking into the nix failure now. If it takes me longer than 30min I'll probably just remove the nix build.

mstange:
Oh I think the nix failure was just because you didn't update the lockfile. So if you run cargo check or so then it'll update Cargo.lock, and you need to include that in the PR.

animalan:
Oops! Thanks! -- fixed :D

PR #808, opened 2026-05-06
Switch back from aws-lc to ring

mstange:
The update to reqwest 0.13 in #801 changed the rustls provider from ring to aws-lc. This PR makes us use ring again.

We're already pulling in ring via symsrv's reqwest 0.12 dependency.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -164,28 +164,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c08606f8c3cbf4ce6ec8e28fb0014a2c086708fe954eaa885384a6165172e7e8"
 
 [[package]]
-name = "aws-lc-rs"
-version = "1.16.3"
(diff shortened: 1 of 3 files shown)


mstange:
I'm doing this because the samply "toolchain build" in Firefox CI was failing; it's building samply against glibc 2.19, which doesn't have `getentropy`.

The error is:

```
warning: aws-lc-sys@0.40.0: /builds/worker/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/aws-lc-sys-0.40.0/aws-lc/crypto/rand_extra/getentropy.c:35:9: error: call to undeclared function 'getentropy'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
warning: aws-lc-sys@0.40.0:    35 |     if (getentropy(out, request_chunk) != 0) {
warning: aws-lc-sys@0.40.0:       |         ^
warning: aws-lc-sys@0.40.0: 1 error generated.
```

PR #809, opened 2026-05-07
Set Mach absolute time platform-specific reference timestamp on macOS

not-matthias:
I ran into an issue on MacOS, when trying to reconstruct the monotonic timer value of a specific sample (since the reference timestamp was discarded before). My use case is extracting specific samples based on a monotonic time range. 

By saving it inside the profile.json, we can reconstruct it again by adding the reference timestamp.


not-matthias:
(sorry about the noise on this one, accidentally dropped my commit while trying to update the branch, see new PR here: https://github.com/mstange/samply/pull/810)

PR #810, opened 2026-05-07
Set Mach absolute time platform-specific reference timestamp on macOS

not-matthias:
I ran into an issue on MacOS, when trying to reconstruct the monotonic timer value of a specific sample (since the reference timestamp was discarded before). My use case is extracting specific samples based on a monotonic time range.

By saving it inside the profile.json, we can reconstruct it again by adding the reference timestamp.

### DIFF
--- a/samply/src/mac/sampler.rs
+++ b/samply/src/mac/sampler.rs
@@ -5,7 +5,8 @@ use std::{mem, thread};
 
 use crossbeam_channel::Receiver;
 use fxprof_processed_profile::{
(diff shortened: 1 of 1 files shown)


mstange (approved):
Oh I'm surprised we weren't doing this already. Thanks!

For some reason this PR doesn't have an "Update branch" button so I can't merge it until you push a rebased / merged version. Maybe you didn't have the "allow edits by maintainer" checkbox checked?

PR #827, opened 2026-06-01
fix(mac): don't SIGKILL Apple toolchain binaries under `samply record`

not-matthias:
Ran into this when profiling Go binaries with `-ldflags=-s=false -w=false` which invokes dsymutils.


samply-mac-preload/src/lib.rs
```diff
@@ -35,7 +35,52 @@ static __SETUP_SAMPLY_CONNECTION: unsafe extern "C" fn() = {
...
+    const CS_OPS_STATUS: u32 = 0;
+    const CS_PLATFORM_BINARY: u32 = 0x0400_0000;
+    extern "C" {
+        fn csops(
+            pid: libc::pid_t,
+            ops: u32,
+            useraddr: *mut libc::c_void,
+            usersize: libc::size_t,
+        ) -> libc::c_int;
+    }
```
mstange, line 64: This seems to be a private API. Is this really the only way to detect this? I'd rather not rely on private APIs.

mstange (changes requested):
I'm actually surprised that `DYLD_INSERT_LIBRARIES` isn't blocked for dsymutil. Are there any other shipping macOS binaries that allow `DYLD_INSERT_LIBRARIES` but don't allow task port sending?

Do you know if the kernel code which triggers the SIGKILL is public? Maybe it contains hints for other detection mechanisms that don't require private APIs.

PR #840, opened 2026-07-22
Read dyld subcache names from the cache header

### LINKED ISSUE #839: macOS 26: cache-only dylibs fail to symbolicate ("Incorrect number of SubCaches")
On macOS 26 the dyld subcache files are no longer all numeric:

```
dyld_shared_cache_arm64e
dyld_shared_cache_arm64e.01
dyld_shared_cache_arm64e.02.dylddata
dyld_shared_cache_arm64e.03.dyldreadonly
dyld_shared_cache_arm64e.04.dyldlinkedit
dyld_shared_cache_arm64e.05
...
```

`advance_to_next_subcache` in samply-symbols probes `.{i}` / `.{i:02}` and stops when both
miss, so it loads `.01`, misses `.02.dylddata`, and `DyldCache::parse` rejects the
incomplete set with "Incorrect number of SubCaches".

As a result, any dylib that only exists in the cache (libsystem_c, libsystem_malloc,
Foundation, ...) gets no symbols. libsystem_kernel still has an on-disk file, so syscall frames
still work.

Fix: get the suffix list from the cache header via `object`'s
`DyldCache::subcache_suffixes` instead of probing. It handles the V1 (numeric) and V2
(suffix string) formats plus `.symbols`, and I've confirmed the current `object` version
parses this cache fine when given those suffixes.

Repro on any Tahoe machine:

```rust
SymbolManager::with_config(SymbolManagerConfig::default())
    .load_symbol_map_for_binary_at_path(Path::new("/usr/lib/system/libsystem_c.dylib"), None)
    .await
// Err: open_file helper callback for file
//   LocalFile("/System/Library/dyld/dyld_shared_cache_x86_64")
//   returned error: No such file or directory (os error 2)
```

Aside 1: that error is misleading: `load_binary_for_dyld_cache_image` keeps only the last
candidate's error, so the arm64e parse failure is replaced by a complaint about an x86_64
cache path that doesn't exist on any arm64 Mac. Aggregating candidate errors like
`load_symbol_map` does (`all_errors`) would have pointed straight at the bug.

Aside 2: a wrong-arch cache that *does* parse but fails the UUID check overwrites earlier
errors the same way, so even with this bug fixed, a debug-id mismatch on the last
candidate can mask a real failure on an earlier one.

I saw this on samply 0.13.1 originally but the code in main is equivalent.

I have a PR ready if you want me to fix this!

LalitMaganti:
macOS 26 uses suffixes such as `.02.dylddata` and
`.03.dyldreadonly`, which cannot be inferred from the subcache index. The
old probing loop stopped after finding `.01`, causing dyld cache parsing
to fail with an incorrect subcache count.

Use `DyldCache::subcache_suffixes` to load the names recorded in the
header. This also handles the numeric V1 format and the optional
`.symbols` cache. Treat every listed subcache as required and report the
specific file when loading fails.

Add regression tests for V1 and V2 cache layouts and missing subcaches.

Fixes: https://github.com/mstange/samply/issues/839

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2663,6 +2663,7 @@ dependencies = [
  "samply-object",
  "scala-native-demangle",
  "srcsrv",
+ "tempfile",
  "thiserror 2.0.18",
  "uuid",
  "yoke",
--- a/samply-symbols/Cargo.toml
+++ b/samply-symbols/Cargo.toml
@@ -62,3 +62,4 @@ crc32fast = "1.4.2"
 [dev-dependencies]
 memmap2 = "0.9.9"
 anyhow = "1"
+tempfile = "3"
--- a/samply-symbols/src/macho.rs
+++ b/samply-symbols/src/macho.rs
@@ -575,16 +575,9 @@ fn read_uleb128(mut bytes: &[u8]) -> Option<(u64, &[u8])> {
     None
 }
 
-/// State machine for loading a dyld shared cache and its numeric/`.symbols`
-/// subcaches.
+/// State machine for loading a dyld shared cache and its subcaches.
 pub struct DyldCacheLoad<FT: FileTypes> {
     state: DyldCacheLoadState<FT>,
-}
-
-#[derive(Clone, Copy, PartialEq, Eq)]
-enum SuffixAttempt {
-    Plain,
-    ZeroPadded,
 }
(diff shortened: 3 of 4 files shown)


mstange:
Thanks!

PR #853, opened 2026-08-04
Update object dependency

philipc:
Things to note:

- `macho_arch_name_for_cpu_type` now ignores the capability bits of the subtype, which matters for arm64e.

- `Object::exports` now obtains exports from more places. For example, for Mach-O it can obtain them from LC_DYLD_EXPORTS_TRIE, LC_DYLD_INFO, or LC_DYSYMTAB (in order of preference).

- The SHT_NOBITS change in `drop_phdr` is a bug fix, but I'm not sure if this was ever hit in practice.

- Changed `drop_phdr` to use `object::write::elf::Encoder` which allows it to be more concise.

---
The only testing I've done is `cargo test`; I'm creating the PR because this `object` release had lots of breaking changes.

### DIFF
--- a/samply/src/linux_shared/object_rewriter.rs
+++ b/samply/src/linux_shared/object_rewriter.rs
@@ -1,7 +1,7 @@
 use std::error::Error;
 
 use object::read::elf::{FileHeader, SectionHeader};
-use object::{elf, Endianness, U16, U32, U64};
+use object::{elf, Endianness};
 
 /// Returns a `Vec<u8>` with ELF data of an equivalent image with no program
 /// headers and fixed section offsets.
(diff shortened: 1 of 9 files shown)


samply/src/linux_shared/object_rewriter.rs
```diff
@@ -85,70 +85,26 @@ pub fn drop_phdr<Elf: FileHeader<Endian = Endianness>>(
     out_data.reserve(reserved_len);
 
     // We're done reserving, start writing.
+    let encoder = object::write::elf::Encoder::new(endian, is_64, in_header.e_machine(endian));
 
     // Write the ELF header.
     {
         // We copy all information from the original header, except the header size + version,
         // the section table offset, and the program header information (set to offset 0 + num 0).
-        // This code would be simpler if the elf::FileHeader trait had setters, and not just getters.
```
mstange, line 93: Hah! I had forgotten about this code. It doesn't get used much these days. But thanks, this looks much simpler!

mstange (approved):
Thank you!

